PR #2, opened 2020-12-07
Fix import paths, missing vendor files in repo move.

wlynch:
This change primarily focuses on 2 things:

1. Renames module path from github.com/tektoncd/experimental/results to
github.com/tektoncd/results.
2. Adds missing vendored plumbing scripts to fix broken CI tests. vendor directory was copied verbatim from experimental/vendor for compatibility.

Note on YAML indentation: This was autogenerated by prettier, though yamllint accepts both forms. Sticking with prettier output, since this is likely what we'll standardize on in the future.

PR #3, opened 2020-12-08
[results] Implement v1alpha2 ListResults.

wlynch:
Largely caries over from v1alpha1 impl, refactoring common libraries.
Notable changes:

- Refactors pagination batch sizing and token encoding into its own library
to share between versions.
- Moves internal pagination proto out of the API package into its own library.
- Refactors CEL environment creation / parsing into its own library to
share between versions.
- Adds additional test coverage for refactored libraries.
- Adds test db logger which prints out any SQL statements made during
tests to t.Log.

This was originally created in https://github.com/tektoncd/experimental/pull/683, but moved here for the new results repo.


wlynch:
/hold

Depends on https://github.com/tektoncd/results/pull/2

dibyom (commented):
/approve
/lgtm

PR #4, opened 2020-12-09
[results] Fix lint errors.

wlynch:
Fixes a variety of linting errors that are blocking presubmit (not
entirely sure why things just started to break).

- Adds missing explicit error checking.
- Removes unused variables/functions.
- Uses explicit struct field names.
- Removes use of deprecated functions (where applicable).
- Removes mutex copying issue due to proto dereferencing.

This is a copy of https://github.com/tektoncd/experimental/pull/680


dibyom (approved):
/lgtm

PR #10, opened 2020-12-10
Adds the UpdateTime field to the Result message

XinruZhang:
Adds the UpdateTime field.

We can use the field to generate the etag for a result.


pkg/api/server/v1alpha2/results.go
```diff
@@ -26,6 +27,8 @@ func (s *Server) CreateResult(ctx context.Context, req *pb.CreateResultRequest)
 
 	// Populate Result with server provided fields.
 	r.Id = uid()
+	r.CreatedTime = timestamppb.Now()
+	r.UpdatedTime = timestamppb.Now()
 
 	store, err := result.ToStorage(parent, name, req.GetResult())
```
wlynch, line 33: As is, I think this is going to strip out the timestamp in the ToStorage function, since we don't actually store this data anywhere in the model.

Let's use a [Clock](https://pkg.go.dev/github.com/jonboulle/clockwork) to make the timestamps deterministic and test to make sure they can be successfully retrieved.

wlynch (changes requested):

pkg/api/server/v1alpha2/results.go
XinruZhang: Great! Done :)

pkg/api/server/db/model.go
```diff
@@ -6,6 +6,7 @@ type Result struct {
 	Parent string `gorm:"primaryKey;index:results_by_name,priority:1"`
 	ID     string `gorm:"primaryKey"`
 	Name   string `gorm:"index:results_by_name,priority:2"`
+	Data   []byte
```
wlynch, line 9: I think we might want to inline these fields directly, so that results can be sorted / indexed by the timestamp values. Unlike records, results don't contain arbitrary user blobs, so it doesn't make as much sense to store the API proto here.

wlynch (changes requested):

pkg/api/server/db/model.go
XinruZhang: Yeah exactly, I'm working on this

pkg/api/server/v1alpha2/result/result_test.go
```diff
@@ -81,23 +84,26 @@ func TestParseName(t *testing.T) {
 }
 
 func TestToStorage(t *testing.T) {
-	got, err := ToStorage("foo", "bar", &pb.Result{
-		Name: "foo/results/bar",
-		Id:   "a",
+	apiResult := &pb.Result{
```
wlynch, line 87: nit: this isn't being reused, so let's just leave it inlined like it was before.

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -49,7 +51,11 @@ func TestCreateResult(t *testing.T) {
 		}
 		want := proto.Clone(req.GetResult()).(*pb.Result)
 		want.Id = fmt.Sprint(lastID)
-		if diff := cmp.Diff(got, want, protocmp.Transform()); diff != "" {
+		opts := cmp.Options{
+			protocmp.Transform(),
+			protocmp.IgnoreFields(&pb.Result{}, "created_time", "updated_time"),
```
wlynch, line 56: I wouldn't ignore the fields here - we're not testing that these fields are actually set (this would be okay for watcher tests).

wlynch (changes requested):
Looking good! Just small things left.

pkg/api/server/v1alpha2/result/result_test.go
XinruZhang: Great! Done

pkg/api/server/v1alpha2/results_test.go
XinruZhang: That's right! Done.

wlynch (approved):
/lgtm

PR #11, opened 2020-12-10
[v1alpha2] Add Create/GetRecord.

### LINKED ISSUE #6: [v1alpha2] CreateRecord
Tracking issue for porting [Create Record](https://github.com/tektoncd/results/blob/319047f8d1c3355d305076cbbe426c56198d9a02/proto/v1alpha2/api.proto#L48-L53) (formerly known as Executions) behavior from [v1alpha1](https://github.com/tektoncd/results/blob/319047f8d1c3355d305076cbbe426c56198d9a02/proto/v1alpha1/api.proto#L65-L68) to [v1alpha2](https://github.com/tektoncd/results/blob/319047f8d1c3355d305076cbbe426c56198d9a02/proto/v1alpha2/resources.proto#L41-L55).

wlynch:
- Implements CreateRecord and GetRecord v1alpha2 RPC methods
- Enforces foreign key constraints for local sqlite3 testing.
- Updates v1alpha1 to also write Results due to foreign key constraints
(since we are getting rid of this soon, this was not done in a
transactionally safe way).

Fixes #6, #7

### DIFF
--- a/pkg/api/server/test/db.go
+++ b/pkg/api/server/test/db.go
@@ -60,6 +60,10 @@ func NewDB(t *testing.T) *gorm.DB {
 		t.Fatalf("failed to open the results.db: %v", err)
 	}
 
+	// Enable foreign key support. Only needed for sqlite instance we use for
+	// tests.
+	gdb.Exec("PRAGMA foreign_keys = ON;")
+
 	return gdb
 }
 
--- a/pkg/api/server/v1alpha2/records_test.go
+++ b/pkg/api/server/v1alpha2/records_test.go
@@ -1,0 +1,206 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/results/pkg/api/server/test"
+	recordutil "github.com/tektoncd/results/pkg/api/server/v1alpha2/record"
+	resultutil "github.com/tektoncd/results/pkg/api/server/v1alpha2/result"
+
+	pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/testing/protocmp"
+)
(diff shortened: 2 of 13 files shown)


pkg/api/server/v1alpha2/records_test.go
```diff
@@ -0,0 +1,206 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/results/pkg/api/server/test"
+	recordutil "github.com/tektoncd/results/pkg/api/server/v1alpha2/record"
```
dibyom, line 10: do we have to alias these imports or can we also just go record.FormatName etc?

pkg/api/server/test/db.go
```diff
@@ -60,6 +60,9 @@ func NewDB(t *testing.T) *gorm.DB {
 		t.Fatalf("failed to open the results.db: %v", err)
 	}
 
+	// Enable foreign key support
+	gdb.Exec("PRAGMA foreign_keys = ON;")
```
dibyom, line 64: does this need to go into the SQL file too?

dibyom (approved):

pkg/api/server/v1alpha2/records_test.go
wlynch: We don't have to, but I did this since we're using the "record" identifier elsewhere and I wanted to make a clearer distinction.

pkg/api/server/test/db.go
wlynch: No. This is something that's only needed for sqlite, so it only needs to be applied for tests (where as the .sql file is used by both sqlite and mysql).

wlynch: FWIW I'm not a big fan of this split usage of databases for tests vs deploy, but AFAICT there's no easy way to spin up a local mysql instance so it's the best we have at the moment for local testing with a real DB.

dibyom: yeah that's unfortunate. I saw https://github.com/proullon/ramsql though that seems unmaintained.

PR #14, opened 2020-12-14
Implement v1alpha2 UpdateResult

XinruZhang:
The logic is basically the same as the v1alpha1 version. The only one
difference is there isn't `executions` field in the Result definition,
so the UpdateResult function won't tackle that field anymore, instead,
UpdateRecord will take charge.

### DIFF
--- a/pkg/api/server/db/model.go
+++ b/pkg/api/server/db/model.go
@@ -2,6 +2,9 @@
 package db
 
 import (
+	"database/sql/driver"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 )
@@ -13,6 +16,7 @@ type Result struct {
 	Name        string `gorm:"index:results_by_name,priority:2"`
 	CreatedTime time.Time
 	UpdatedTime time.Time
+	Annotations dbjson
 }
 
 func (r Result) String() string {
@@ -27,4 +31,27 @@ type Record struct {
 	ResultName string `gorm:"index:records_by_name,priority:2"`
 	Name       string `gorm:"index:records_by_name,priority:3"`
 	Data       []byte
+}
+
+type dbjson json.RawMessage
+
+// Scan scan value into Jsonb, implements sql.Scanner interface
+func (j *dbjson) Scan(value interface{}) error {
+	bytes, ok := value.([]byte)
+	if !ok {
+		return errors.New(fmt.Sprint("Failed to unmarshal JSONB value:", value))
+	}
+
+	result := json.RawMessage{}
+	err := json.Unmarshal(bytes, &result)
+	*j = dbjson(result)
+	return err
+}
+
+// Value return json value, implement driver.Valuer interface
+func (j dbjson) Value() (driver.Value, error) {
+	if len(j) == 0 {
+		return nil, nil
+	}
+	return json.RawMessage(j).MarshalJSON()
 }
--- a/pkg/api/server/v1alpha2/result/result.go
+++ b/pkg/api/server/v1alpha2/result/result.go
@@ -2,6 +2,7 @@
 package result
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"regexp"
@@ -12,6 +13,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
+	"gorm.io/gorm"
 )
 
 var (
@@ -31,9 +33,12 @@ func ParseName(raw string) (parent, name string, err error) {
 
 // ToStorage converts an API Result into its corresponding database storage
 // equivalent.
-// parent,name should be the name parts (e.g. not containing "/results/").
 func ToStorage(r *pb.Result) (*db.Result, error) {
 	parent, name, err := ParseName(r.GetName())
+	if err != nil {
+		return nil, err
+	}
+	ann, err := json.Marshal(r.GetAnnotations())
 	if err != nil {
 		return nil, err
 	}
@@ -43,6 +48,7 @@ func ToStorage(r *pb.Result) (*db.Result, error) {
 		Name:        name,
 		UpdatedTime: r.UpdatedTime.AsTime(),
 		CreatedTime: r.CreatedTime.AsTime(),
+		Annotations: ann,
 	}
 	return result, nil
 }
@@ -50,9 +56,14 @@ func ToStorage(r *pb.Result) (*db.Result, error) {
 // ToAPI converts a database storage Result into its corresponding API
 // equivalent.
 func ToAPI(r *db.Result) *pb.Result {
+	ann := map[string]string{}
+	if err := json.Unmarshal(r.Annotations, &ann); err != nil {
+		ann = nil
+	}
 	return &pb.Result{
 		Name:        fmt.Sprintf("%s/results/%s", r.Parent, r.Name),
 		Id:          r.ID,
+		Annotations: ann,
 		CreatedTime: timestamppb.New(r.CreatedTime),
 		UpdatedTime: timestamppb.New(r.UpdatedTime),
 	}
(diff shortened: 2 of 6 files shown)


pkg/api/server/v1alpha2/result/result.go
```diff
@@ -80,3 +91,23 @@ func Match(r *pb.Result, prg cel.Program) (bool, error) {
...
+	if err := gdb.Where("parent = ?", parent).Where("name = ?", name).Find(&results).Error; err != nil {
+		log.Printf("failed to query on database: %v", err)
+		return nil, fmt.Errorf("failed to query on a result: %w", err)
+	}
+	if len(results) == 0 {
+		return nil, status.Error(codes.NotFound, "result not found")
+	}
+	if len(results) > 1 {
+		log.Println("Warning: multiple rows found")
+	}
```
wlynch, line 111: You can simplify this a bit by using the `First` operator - see https://gorm.io/docs/query.html#Retrieving-objects-with-primary-key

You might find https://github.com/tektoncd/results/pull/11/files#diff-fc57ccd1ec8e2040c7ebcf3b43f29ccdd06e4c7cda1a66bba8875c6f41902a95R61-R67 as a useful example.

pkg/api/server/v1alpha2/result/result.go
```diff
@@ -80,3 +91,23 @@ func Match(r *pb.Result, prg cel.Program) (bool, error) {
...
 }
+
+// GetResultByName is the helper function to get a Result by its parent and name
+func GetResultByName(gdb *gorm.DB, name string) (*pb.Result, error) {
+	parent, name, err := ParseName(name)
+	if err != nil {
+		return nil, err
+	}
+	var results []*db.Result = []*db.Result{}
+	if err := gdb.Where("parent = ?", parent).Where("name = ?", name).Find(&results).Error; err != nil {
```
wlynch, line 102: Don't forget to wrap the error (this should also cover NotFound errors)

pkg/api/server/v1alpha2/result/result.go
```diff
@@ -31,28 +33,37 @@ func ParseName(raw string) (parent, name string, err error) {
...
 	}
 	return result, nil
 }
 
 // ToAPI converts a database storage Result into its corresponding API
 // equivalent.
 func ToAPI(r *db.Result) *pb.Result {
+	ann := map[string]string{}
+	if err := json.Unmarshal(r.Annotations, &ann); err != nil {
+		ann = nil
```
wlynch, line 61: We should return the error here.

pkg/api/server/db/model.go
```diff
@@ -28,3 +32,26 @@ type Record struct {
 	Name       string `gorm:"index:records_by_name,priority:3"`
 	Data       []byte
 }
+
+type dbjson json.RawMessage
+
+// Scan scan value into Jsonb, implements sql.Scanner interface
```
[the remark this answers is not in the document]
wlynch: Reading this some more - how is this invoked, and how does this relate to the Unmarshal/Marshal in result.go?

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -147,6 +149,112 @@ func TestGetResult(t *testing.T) {
...
+	srv, err := New(test.NewDB(t))
+	if err != nil {
+		t.Fatalf("failed to create server: %v", err)
+	}
+	ctx := context.Background()
+
+	tt := []struct {
+		name      string
+		parent    string
+		in        *pb.Result
```
wlynch, line 162: `in` isn't used meaningfully - AFAICT only to set unique names. We can generate pseudo-random names from the loop index instead.

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -147,6 +149,112 @@ func TestGetResult(t *testing.T) {
...
+		fieldmask *field_mask.FieldMask
+		update    *pb.Result
+		expect    *pb.Result
+		errcode   codes.Code
+	}{
+		{
+			name:   "Test no Mask",
+			in:     &pb.Result{Name: "foo/results/bar-001"},
+			update: &pb.Result{Annotations: map[string]string{"foo": "bar"}},
+			expect: &pb.Result{Annotations: map[string]string{"foo": "bar"}},
```
wlynch, line 172: It wasn't intuitive on first pass that that this was a partial response. We should add a comment above explaining the behavior.

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -147,6 +149,112 @@ func TestGetResult(t *testing.T) {
...
+			}
+			// Update the created taskrun
+			r, err = srv.UpdateResult(ctx, &pb.UpdateResultRequest{Result: tc.update, Name: tc.update.GetName(), UpdateMask: tc.fieldmask})
+			if err != nil {
+				if status.Code(err) != tc.errcode {
+					t.Fatalf("expected error: %v, got %v", tc.errcode, status.Code(err))
+				} else {
+					return
+				}
+				t.Fatalf("could not update taskrun: %v, %v", err, status.Code(err))
```
wlynch, line 237: This might be easier to read if we simplify the logic a bit:

```suggestion
				if status.Code(err) == tc.errcode {
					continue
				}
				t.Fatalf("UpdateResult()=(%v, %v); want %v", r, err, tc.errcode)
```

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -147,6 +149,112 @@ func TestGetResult(t *testing.T) {
...
+			}
+
+			// Expected results should always match the created result.
+			if tc.expect != nil {
+				tc.expect.Name = tc.in.GetName()
+				tc.expect.Id = r.GetId()
+				tc.expect.CreatedTime = r.CreatedTime
+				tc.expect.UpdatedTime = r.UpdatedTime
```
wlynch, line 245: Let's set an explicit time here - if we didn't actually update the time, this would still pass as is. Don't forget to increment the fake clock!

wlynch (changes requested):
Thanks for the PR!

I think we should break this up into 2-3 different pieces:
1) Add support for Result Annotations.
2) Add support for Updating results
3*) Add support for UpdateResult field masks (*I'm not actually convinced this is useful yet, if all you can update is 1 field - we should probably remove the field mask field until we have a clearer use case)

That said, initial comments below.

pkg/api/server/db/model.go
```diff
@@ -28,3 +32,26 @@ type Record struct {
...
+func (j *dbjson) Scan(value interface{}) error {
+	bytes, ok := value.([]byte)
+	if !ok {
+		return errors.New(fmt.Sprint("Failed to unmarshal JSONB value:", value))
+	}
+
+	result := json.RawMessage{}
+	err := json.Unmarshal(bytes, &result)
+	*j = dbjson(result)
```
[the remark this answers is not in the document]
XinruZhang: Done

pkg/api/server/db/model.go
```diff
@@ -28,3 +32,26 @@ type Record struct {
...
+
+// Scan scan value into Jsonb, implements sql.Scanner interface
+func (j *dbjson) Scan(value interface{}) error {
+	bytes, ok := value.([]byte)
+	if !ok {
+		return errors.New(fmt.Sprint("Failed to unmarshal JSONB value:", value))
+	}
+
+	result := json.RawMessage{}
+	err := json.Unmarshal(bytes, &result)
```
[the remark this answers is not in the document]
XinruZhang: Done

pkg/api/server/db/model.go
```diff
@@ -28,3 +32,26 @@ type Record struct {
...
 	Data       []byte
 }
+
+type dbjson json.RawMessage
+
+// Scan scan value into Jsonb, implements sql.Scanner interface
+func (j *dbjson) Scan(value interface{}) error {
+	bytes, ok := value.([]byte)
+	if !ok {
+		return errors.New(fmt.Sprint("Failed to unmarshal JSONB value:", value))
```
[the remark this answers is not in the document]
XinruZhang: Done

pkg/api/server/db/model.go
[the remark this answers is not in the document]
XinruZhang: I guess I didn't define this type correctly, I should define a map[string]string type directly. I was concentrating on how to save that field into the database, then serialized json string comes into my mind.

pkg/api/server/v1alpha2/result/result.go
XinruZhang: Done

pkg/api/server/v1alpha2/results_test.go
XinruZhang: Done

pkg/api/server/v1alpha2/results_test.go
XinruZhang: No masks, so done.

pkg/api/server/v1alpha2/results_test.go
XinruZhang: Done

wlynch:
/hold

Looks like a lot of changes in https://github.com/tektoncd/results/pull/19 still remain here, so holding until that is resolved.

XinruZhang:
move to another PR #25

PR #19, opened 2020-12-15
Add a new column "annotations" to the table result

XinruZhang:
Adds a new column "Annotations" to the result table schema.

Defines a new type "Annotations" and implements sql.Scanner and driver.Valuer interface to let gorm know how to receive and save that field from/to database.

Get this new field set in the "result" model when testing CreateResult, ToStorage, and ToAPI.

### DIFF
--- a/pkg/api/server/db/model.go
+++ b/pkg/api/server/db/model.go
@@ -2,6 +2,9 @@
 package db
 
 import (
+	"database/sql/driver"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 )
@@ -13,6 +16,7 @@ type Result struct {
 	Name        string `gorm:"index:results_by_name,priority:2"`
 	CreatedTime time.Time
 	UpdatedTime time.Time
+	Annotations Annotations
 }
 
 func (r Result) String() string {
@@ -32,4 +36,34 @@ type Record struct {
 	ID   string `gorm:"primaryKey"`
 	Name string `gorm:"index:records_by_name,priority:3"`
 	Data []byte
+}
+
+// Annotations is a custom-defined type of a gorm model field.
+type Annotations map[string]string
+
+// Scan resolves serialized data read from database into an Annotation.
+// This implements the sql.Scanner interface.
+func (ann *Annotations) Scan(value interface{}) error {
+	if ann == nil {
+		return errors.New("the annotation pointer mustn't be nil")
+	}
+	bytes, ok := value.([]byte)
+	if !ok {
+		return fmt.Errorf("wanted []byte, got %T: %+v", value, value)
+	}
+	if err := json.Unmarshal(bytes, ann); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Value returns the value of Annotations for database driver. This implements driver.Valuer.
+// gorm uses this function to convert a database model's Annotation field into a type that gorm
+// driver can write into the database.
+func (ann Annotations) Value() (driver.Value, error) {
+	bytes, err := json.Marshal(ann)
+	if err != nil {
+		return nil, err
+	}
+	return bytes, nil
 }
--- a/pkg/api/server/db/model_test.go
+++ b/pkg/api/server/db/model_test.go
@@ -1,0 +1,51 @@
+package db
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestAnnotationsScan(t *testing.T) {
+	v := make(Annotations)
+	v["foo"] = "bar"
+
+	bytes, err := json.Marshal(v)
+	if err != nil {
+		t.Fatalf("json.Marshal(): %v", err)
+	}
+
+	var ann *Annotations
+	if err := ann.Scan(bytes); err == nil {
+		t.Error("annotation pointer must not be nil, expected error")
+	}
+
+	ann = &Annotations{}
+	if err := ann.Scan(bytes); err != nil {
+		t.Fatalf("failed to scan data from database: %v", err)
+	}
+
+	if diff := cmp.Diff(*ann, v); diff != "" {
+		t.Errorf("-want, +got: %s", diff)
+	}
+}
+
+func TestAnnotationsValue(t *testing.T) {
+	v := make(Annotations)
+	v["foo"] = "bar"
+
+	bytes, err := json.Marshal(v)
+	if err != nil {
+		t.Fatalf("json.Marshal(): %v", err)
+	}
+
+	annv, err := v.Value()
+	if err != nil {
+		t.Fatalf("Value(): %v", err)
+	}
+
+	if diff := cmp.Diff(annv, bytes); diff != "" {
+		t.Errorf("-want, +got: %s", diff)
+	}
+}
--- a/pkg/api/server/v1alpha2/result/result.go
+++ b/pkg/api/server/v1alpha2/result/result.go
@@ -49,6 +49,7 @@ func ToStorage(r *pb.Result) (*db.Result, error) {
 		Name:        name,
 		UpdatedTime: r.UpdatedTime.AsTime(),
 		CreatedTime: r.CreatedTime.AsTime(),
+		Annotations: r.Annotations,
 	}
 	return result, nil
 }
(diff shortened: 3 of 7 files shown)


pkg/api/server/db/model.go
```diff
@@ -28,3 +32,32 @@ type Record struct {
...
 	Data       []byte
 }
+
+// Annotations is a custom-defined type of a gorm model field.
+type Annotations map[string]string
+
+// Scan implements sql.Scanner interface, to let gorm know how to resolve the corresponding data from database.
+func (ann *Annotations) Scan(value interface{}) error {
+	if ann == nil {
+		return errors.New("nil pointer")
```
wlynch, line 42: If annotations is unset in the proto, will this cause an error? If so, I think we should allow that.

Let's make sure to add a test to verify the behavior.

pkg/api/server/db/model.go
```diff
@@ -28,3 +32,32 @@ type Record struct {
...
+	}
+	bytes, ok := value.([]byte)
+	if !ok {
+		return errors.New(fmt.Sprint("failed to obtain a byte array form database:", value))
```
wlynch, line 46: It might be useful to also include the type we don't know what this might be if this fails.

```suggestion
		return errors.New(fmt.Sprintf("wanted []byte, got %T: %+v", value, value))
```

pkg/api/server/db/model.go
```diff
@@ -28,3 +32,32 @@ type Record struct {
...
+	}
+	if err := json.Unmarshal(bytes, ann); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Value returns the value of Annotations for database driver, implement driver.Valuer interface.
+// gorm uses this function to convert a database model's Annotation field into a type that gorm
+// dirver can write into the database.
```
wlynch, line 56:
```suggestion
// Value returns the value of Annotations for database driver. This implements driver.Valuer.
// gorm uses this function to convert a database model's Annotation field into a type that gorm
// driver can write into the database.
```

pkg/api/server/db/model.go
```diff
@@ -28,3 +32,32 @@ type Record struct {
 	Name       string `gorm:"index:records_by_name,priority:3"`
 	Data       []byte
 }
+
+// Annotations is a custom-defined type of a gorm model field.
+type Annotations map[string]string
+
+// Scan implements sql.Scanner interface, to let gorm know how to resolve the corresponding data from database.
```
wlynch, line 39:
```suggestion
// Scan resolves serialized data read from database into an Annotation.
// This implements the sql.Scanner interface.
```

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -42,14 +42,19 @@ func TestCreateResult(t *testing.T) {
...
+			Name:        "foo/results/bar",
+			Annotations: map[string]string{"foo": "bar"},
 		},
 	}
 	t.Run("success", func(t *testing.T) {
 		got, err := srv.CreateResult(ctx, req)
 		if err != nil {
 			t.Fatalf("could not create result: %v", err)
 		}
+		got, err = srv.GetResult(ctx, &pb.GetResultRequest{Name: got.GetName()})
```
wlynch, line 54: Since this is likely going to be a common use case, let's also add a test filtering a result by the annotation data for completeness.

wlynch (changes requested):
Looking good! :D

pkg/api/server/db/model.go
XinruZhang: Done

pkg/api/server/db/model.go
XinruZhang: Done

pkg/api/server/v1alpha2/result/result.go
```diff
@@ -49,19 +49,21 @@ func ToStorage(r *pb.Result) (*db.Result, error) {
...
-func ToAPI(r *db.Result) *pb.Result {
+func ToAPI(r *db.Result) (*pb.Result, error) {
 	return &pb.Result{
 		Name:        FormatName(r.Parent, r.Name),
 		Id:          r.ID,
 		CreatedTime: timestamppb.New(r.CreatedTime),
 		UpdatedTime: timestamppb.New(r.UpdatedTime),
-	}
+		Annotations: r.Annotations,
+	}, nil
```
wlynch, line 66: We're not using error meaningfully yet - let's remove for now.

pkg/api/server/db/model_test.go
```diff
@@ -0,0 +1,32 @@
+package db
+
+import (
+	"testing"
+
+	"github.com/tektoncd/results/pkg/api/server/db/errors"
+	"github.com/tektoncd/results/pkg/api/server/test"
+)
+
+func TestQueryAnnotation(t *testing.T) {
```
wlynch, line 10: Nice thinking with the test! It might be simpler just to test the Scan/Value functions directly though (e.g. similar to our tests of ToAPI/ToStorage) - as is this basically a simplified version of the results tests below.

wlynch (commented):
Looking good! Just 2 minor comments.

pkg/api/server/db/model_test.go
XinruZhang: Actually this is different from results test because in the results test, every query is wrapped by gorm, the receiver of gorm query is always the db model, so we can't test  the nil annotation pointer behavior:
```go
db.Select("name", "age").Find(&users)
```

However, anyway, I think just to test the Scan/Value functions is a great idea! And this part is done!

pkg/api/server/v1alpha2/result/result.go
XinruZhang: Done :)

pkg/api/server/db/model_test.go
```diff
@@ -0,0 +1,51 @@
...
+	v["foo"] = "bar"
+
+	bytes, err := json.Marshal(v)
+	if err != nil {
+		t.Error(err)
+	}
+
+	var ann *Annotations
+	if err := ann.Scan(bytes); err == nil {
+		t.Errorf("annotation pointer musn't be nil, expect error here.")
```
wlynch, line 21:
```suggestion
		t.Error("annotation pointer must not be nil, expected error")
```

pkg/api/server/db/model_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	ann = &Annotations{}
+	if err := ann.Scan(bytes); err != nil {
+		t.Errorf("fail to scan data from database.")
```
wlynch, line 26:
```suggestion
		t.Fatalf("failed to scan data from database: %v", err)
```

pkg/api/server/db/model_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestAnnotationsScan(t *testing.T) {
+	v := make(Annotations)
+	v["foo"] = "bar"
+
+	bytes, err := json.Marshal(v)
+	if err != nil {
+		t.Error(err)
```
wlynch, line 16:
```suggestion
		t.Fatalf("json.Marshal(): %v", err)
```

pkg/api/server/db/model_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+func TestAnnotationsValue(t *testing.T) {
+	v := make(Annotations)
+	v["foo"] = "bar"
+
+	bytes, err := json.Marshal(v)
+	if err != nil {
+		t.Error(err)
```
wlynch, line 40:
```suggestion
		t.Fatalf("json.Marshal(): %v", err)
```

pkg/api/server/db/model_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	annv, err := v.Value()
+	if err != nil {
+		t.Error(err)
```
wlynch, line 45:
```suggestion
		t.Fatalf("Value(): %v", err)
```

wlynch (commented):
/lgtm
Just some minor formatting / spelling fixes.

pkg/api/server/db/model_test.go
XinruZhang: Done

wlynch (approved):
/lgtm

PR #20, opened 2020-12-15
CreateRecord: store Record data.

wlynch:
This change stores the Record data in the database, which was previously
ignored.

Also introduces an internal protoutil library to help streamline common
marshalling operations in tests. Since these are intended to be test
only, we're liberal with the use of panic to allow for callers to embed
these inline to struct definitions.

Missing part of #6

### DIFF
--- a/pkg/api/server/v1alpha2/record/record_test.go
+++ b/pkg/api/server/v1alpha2/record/record_test.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/tektoncd/results/pkg/api/server/db"
+	"github.com/tektoncd/results/pkg/api/server/internal/protoutil"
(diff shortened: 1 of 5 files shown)


pkg/api/server/v1alpha2/record/record_test.go
```diff
@@ -80,44 +82,108 @@ func TestParseName(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if diff := cmp.Diff(tc.want, got, protocmp.Transform()); diff != "" {
+				t.Errorf("-want,+got: %s", diff)
+			}
+		})
 	}
+
```
XinruZhang, line 186: nit: maybe drop the empty line.

XinruZhang (commented):
Looks good!

pkg/api/server/v1alpha2/record/record_test.go
wlynch: Done.

Issue #5, opened 2020-12-10
[v1alpha2] UpdateResults


wlynch:
Tracking issue for porting [v1alpha1 UpdateResults](https://github.com/tektoncd/results/blob/319047f8d1c3355d305076cbbe426c56198d9a02/pkg/api/server/v1alpha1/server.go#L82) to [v1alpha2](https://github.com/tektoncd/results/blob/319047f8d1c3355d305076cbbe426c56198d9a02/proto/v1alpha2/api.proto#L23-L28).

PR #25, opened 2020-12-17
Implements v1alpha2 UpdateResult

XinruZhang:
In the current version, this api only updates the Annotation field, and the in coming Annotation will overwrite the original one.

Partial work for the milestone #5


XinruZhang:
Based on #19

wlynch:
/hold

Depends on https://github.com/tektoncd/results/pull/19

pkg/api/server/v1alpha2/results.go
```diff
@@ -67,6 +68,34 @@ func (s *Server) GetResult(ctx context.Context, req *pb.GetResultRequest) (*pb.R
 	return result.ToAPI(store), nil
 }
 
+// UpdateResult receives Result and FieldMask from client and uses them to update records in local Sqlite Server.
+func (s Server) UpdateResult(ctx context.Context, req *pb.UpdateResultRequest) (*pb.Result, error) {
```
wlynch, line 72:
```suggestion
func (s *Server) UpdateResult(ctx context.Context, req *pb.UpdateResultRequest) (*pb.Result, error) {
```

pkg/api/server/v1alpha2/results.go
```diff
@@ -67,6 +68,34 @@ func (s *Server) GetResult(ctx context.Context, req *pb.GetResultRequest) (*pb.R
...
+	dbResult, err := result.ToStorage(r)
+	if err != nil {
+		return nil, err
+	}
+	if err := errors.Wrap(tx.Save(&dbResult).Error); err != nil {
+		log.Printf("failed to save result into database: %v", err)
+		tx.Rollback()
+		return nil, err
+	}
+	if err := tx.Commit().Error; err != nil {
```
wlynch, line 92:
```suggestion
	if err := errors.Wrap(tx.Commit().Error); err != nil {
```

pkg/api/server/v1alpha2/results.go
```diff
@@ -67,6 +68,34 @@ func (s *Server) GetResult(ctx context.Context, req *pb.GetResultRequest) (*pb.R
...
+		log.Printf("failed to commit transaction: %v", err)
+		return nil, fmt.Errorf("failed to commit transaction: %v", err)
```
wlynch, line 94: Using fmt.Errorf will remove the RPC status from the error, resulting in UNKNOWN being returned. You'll probably want to do something like:

```
status.Errorf(status.Code(err), "failed to commit transaction: %v", err)
```

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+func TestUpdateResult(t *testing.T) {
+	srv, err := New(test.NewDB(t))
+	if err != nil {
+		t.Fatalf("failed to create server: %v", err)
+	}
+	ctx := context.Background()
+
+	tt := []struct {
+		name    string
+		create  *pb.Result
```
wlynch, line 116: This isn't being used. Remove.

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+		update  *pb.Result
+		expect  *pb.Result
+		errcode codes.Code
```
wlynch, line 119: We should add tests that use this

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+			// create a result for test.
+			r := &pb.Result{Name: fmt.Sprintf("foo/results/bar-%v", idx)}
+			created, err := srv.CreateResult(ctx, &pb.CreateResultRequest{Parent: "foo", Result: r})
+			if err != nil {
+				t.Fatalf("could not create result: %v", err)
+			}
+
+			tc.expect.Name = created.GetName()
+			tc.expect.Id = created.GetId()
+			tc.expect.CreatedTime = created.GetCreatedTime()
```
wlynch, line 142: I think we should use proto.Merge here instead. This will be more resilient to future additions to the proto.

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to create server: %v", err)
+	}
+	ctx := context.Background()
+
+	tt := []struct {
+		name    string
+		create  *pb.Result
+		update  *pb.Result
+		expect  *pb.Result
```
wlynch, line 118: We should add a comment here explaining the behavior of this field - this isn't the Result we expect, but rather the diff from the initially created Result.

pkg/api/server/v1alpha2/server.go
```diff
@@ -47,3 +51,17 @@ func New(db *gorm.DB) (*Server, error) {
 
 	return srv, nil
 }
+
+// GetResultByName is the helper function to get a Result by its parent and name
+func GetResultByName(gdb *gorm.DB, name string) (*pb.Result, error) {
```
wlynch, line 56: This near-identical to GetResult, but is missing the error handling that is in GetResult. We should refactor these together so that they share the same impl but allow for different transaction types to be used. (we should also move this into result.go with the rest of the CRUD operations).

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+		},
+		{
+			name:   "test update with empty result",
+			expect: &pb.Result{},
+		},
+	}
+	for idx, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			// create a result for test.
+			r := &pb.Result{Name: fmt.Sprintf("foo/results/bar-%v", idx)}
```
wlynch, line 134: We can probably just inline this below - the returned `created` can be used everywhere else `r` is used.

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+			}
+
+			tc.expect.UpdatedTime = timestamppb.New(clock.Now())
+
+			got, err := srv.GetResult(ctx, &pb.GetResultRequest{Name: updated.GetName()})
+			if err != nil {
+				t.Fatalf("failed to get result from server: %v", err)
+			}
+
+			if diff := cmp.Diff(tc.expect, got, protocmp.Transform()); diff != "" {
```
wlynch, line 163: Good idea checking that the Update is reflected by the Get, but we should also check that the response from Update is what we expect.

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+			updateRequest := &pb.UpdateResultRequest{Result: tc.update, Name: created.GetName()}
+			updated, err := srv.UpdateResult(ctx, updateRequest)
+			if err != nil {
+				if status.Code(err) == tc.errcode {
+					return
+				}
+				t.Fatalf("UpdateResult()=(%v, %v); want %v", r, err, tc.errcode)
+			}
+
+			tc.expect.UpdatedTime = timestamppb.New(clock.Now())
```
wlynch, line 156: Let's group all of the `expect` field setting in one place.

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+			// foward the time to test if the UpdateTime field is properly updated.
+			fakeClock.Advance(time.Second)
+
+			updateRequest := &pb.UpdateResultRequest{Result: tc.update, Name: created.GetName()}
+			updated, err := srv.UpdateResult(ctx, updateRequest)
+			if err != nil {
+				if status.Code(err) == tc.errcode {
+					return
+				}
+				t.Fatalf("UpdateResult()=(%v, %v); want %v", r, err, tc.errcode)
```
wlynch, line 153:
```suggestion
				t.Fatalf("UpdateResult()=(%v, %v); want %v", updated, err, tc.errcode)
```

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+			}
+
+			tc.expect.Name = created.GetName()
+			tc.expect.Id = created.GetId()
+			tc.expect.CreatedTime = created.GetCreatedTime()
+
+			// foward the time to test if the UpdateTime field is properly updated.
+			fakeClock.Advance(time.Second)
+
+			updateRequest := &pb.UpdateResultRequest{Result: tc.update, Name: created.GetName()}
```
wlynch, line 147: We're not reusing this anywhere, so just inline below.

pkg/api/server/v1alpha2/results.go
```diff
@@ -67,6 +68,34 @@ func (s *Server) GetResult(ctx context.Context, req *pb.GetResultRequest) (*pb.R
...
+func (s Server) UpdateResult(ctx context.Context, req *pb.UpdateResultRequest) (*pb.Result, error) {
+	// Retrieve result from database by name
+	tx := s.db.WithContext(ctx).Begin()
+	r, err := GetResultByName(tx, req.GetName())
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "failed to find a result: %v", err)
+	}
+	r.Annotations = req.GetResult().GetAnnotations()
+	r.UpdatedTime = timestamppb.New(clock.Now())
+
```
wlynch, line 81: Do we want to support etag / other validation here, or want to mark it as TODO for now? (either is fine with me)

pkg/api/server/v1alpha2/results.go
```diff
@@ -67,6 +68,34 @@ func (s *Server) GetResult(ctx context.Context, req *pb.GetResultRequest) (*pb.R
 	return result.ToAPI(store), nil
 }
 
+// UpdateResult receives Result and FieldMask from client and uses them to update records in local Sqlite Server.
+func (s Server) UpdateResult(ctx context.Context, req *pb.UpdateResultRequest) (*pb.Result, error) {
+	// Retrieve result from database by name
+	tx := s.db.WithContext(ctx).Begin()
+	r, err := GetResultByName(tx, req.GetName())
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "failed to find a result: %v", err)
```
wlynch, line 77: What happens if we return here without a rollback? (I'm not sure what gorm's behavior is).

pkg/api/server/v1alpha2/results.go
```diff
@@ -67,6 +68,34 @@ func (s *Server) GetResult(ctx context.Context, req *pb.GetResultRequest) (*pb.R
 	return result.ToAPI(store), nil
 }
 
+// UpdateResult receives Result and FieldMask from client and uses them to update records in local Sqlite Server.
```
wlynch, line 71:
```suggestion
// UpdateResult updates a Result in the database.
```

Figured this is copied from v1alpha1, but we're not updating records here, fieldmask isn't supported yet, and the real deployment doesn't use sqlite. 😅

wlynch: We should also add FieldMask as a TODO (or perhaps since we're not actually using this in a meaningful way, just remove it from the proto altogether).

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -102,6 +104,69 @@ func TestCreateResult(t *testing.T) {
...
+		{
+			name:   "test success",
+			update: &pb.Result{Annotations: map[string]string{"foo": "bar"}},
+			expect: &pb.Result{Annotations: map[string]string{"foo": "bar"}},
+		},
+		{
+			name:   "test update with empty result",
+			expect: &pb.Result{},
+		},
+	}
```
wlynch, line 130: Other good things to test are all of the conditions that return errors.

wlynch (changes requested):

pkg/api/server/v1alpha2/results.go
XinruZhang: Got that, Done

pkg/api/server/v1alpha2/results_test.go
XinruZhang: Done

pkg/api/server/v1alpha2/results.go
XinruZhang: I'm going to open another PR to support ETAG

pkg/api/server/v1alpha2/results_test.go
XinruZhang: Great!

pkg/api/server/v1alpha2/server.go
XinruZhang: Ok!

pkg/api/server/v1alpha2/results.go
XinruZhang: Yeah it's a great point! I'll defer the rollback and commit operation.

pkg/api/server/v1alpha2/results.go
XinruZhang: Okay, I'm going to move the fieldmask field

pkg/api/server/v1alpha2/results_test.go
XinruZhang: Done

pkg/api/server/v1alpha2/results_test.go
XinruZhang: Done

pkg/api/server/v1alpha2/result/result.go
```diff
@@ -73,3 +75,12 @@ func Match(r *pb.Result, prg cel.Program) (bool, error) {
 	}
 	return resultscel.Match(prg, "result", r)
 }
+
+// GetResultByParentName is a helper function to retrieve a Result from database by parent and name
+func GetResultByParentName(gdb *gorm.DB, parent, name string) (*db.Result, error) {
```
wlynch, line 80: I might have typo-ed a previous comment - I think pkg/api/server/v1alpha2/results.go is probably the best spot for this, since this where the other DB operations are.

We should also unexport this so that this isn't publicly visible.

pkg/api/server/v1alpha2/result/result_test.go
```diff
@@ -212,3 +214,21 @@ func TestFormatName(t *testing.T) {
 		t.Errorf("want %s, got %s", want, got)
 	}
 }
+
+func TestGetResultByParentName(t *testing.T) {
```
wlynch, line 218: This is a case where we probably don't need this test, since the GetResult test fully covers the logic contained here, and is a simple enough wrapper around a (soon to be) exported func that we probably doesn't make sense to break out into its own test.

wlynch (changes requested):
Looks good! One last thing.

wlynch (approved):
/lgtm

PR #30, opened 2020-12-22
[v1alpha2] Add UpdateRecord.

wlynch:
Adds initial UpdateRecord behavior - does not include etag and field mask
support (these will be added later).

Minor changes:
- Refactors GetRecord to be transaction-friendly.
- Fixes UpdateRecordRequest to use Record resource.
- Removes UpdateRecordRequest.Name field to be in line with
  https://google.aip.dev/134 guidance.
- Renumbers UpdateRecordRequest fields - while technically an
  incompatible change, since this API is not used yet this is safe to
  do.


wlynch:
/hold

Reviewable, but depends on https://github.com/tektoncd/results/pull/29

dibyom (approved):

PR #31, opened 2020-12-23
watcher/reconciler: Use controller.StartAll for integration test.

wlynch:
Previously the reconciler test was functionaly similar to the
underlying Task/PipelineRun reconciler tests, since the test.Reconcile*
funcs would invoke the Reconcile func directly via the assets.

This changes the behavior of the test to start the controller and
informers, and only interface with the watcher via the Tekton client,
polling for the result annotation to verify that it was successful.
We can rely on the underlying typed reconciler tests to verify that the
result uploading is working as intended.

### DIFF
--- a/pkg/watcher/reconciler/reconciler_test.go
+++ b/pkg/watcher/reconciler/reconciler_test.go
@@ -3,169 +3,106 @@ package reconciler
 import (
 	"context"
 	"testing"
+	"time"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
-	pipelinetest "github.com/tektoncd/pipeline/test"
-	"github.com/tektoncd/results/pkg/watcher/convert"
+	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned/fake"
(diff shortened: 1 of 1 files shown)


pkg/watcher/reconciler/reconciler_test.go
```diff
@@ -3,169 +3,106 @@ package reconciler
...
-	"k8s.io/apimachinery/pkg/types"
+	"knative.dev/pkg/controller"
 )
 
-func TestReconciler(t *testing.T) {
-	testFuncs := map[string]func(*testing.T){
-		"Update PipelineRun to the existed Result": testUpdatePipelineRunToTheExistedResult,
-		"Update TaskRun to the existed Result":     testUpdateTaskRunToTheExistedResult,
-	}
+// TestReconciler starts a full TaskRun + PipelineRun controller and waits for
```
XinruZhang, line 22: Typo here: TestController

XinruZhang (commented):
/lgtm

pkg/watcher/reconciler/reconciler_test.go
wlynch: Good catch! Fixed.

Issue #23, opened 2020-12-16
[v1alpha2] Migrate watcher to v1alpha2 API


wlynch:
Tracking bug for migrating the watcher from the v1alpha1 Results API to v1alpha2.

PR #34, opened 2021-01-05
Migrate TaskRun watcher to v1alpha2 API.

wlynch:
Sorry about the large PR! 🙏  Happy to break this up into smaller pieces if desired - avoided this since I thought the individual pieces may not make sense without the larger picture.

Notable changes:
- TaskRun watcher now uses v1alpha2 API instead of v1alpha1.
- Adds ToProto convert func to help controllers be Record-type agnostic.
- Annotation library now supports Record annotations.
- Creates Results helper client to handle type agnostic Result API
  operations.
- Refactors TaskRun reconciler test:
  - taskRunTest -> env
  - test functions combined into single test sparated by t.Run so that they
  can build off each other (create -> nop -> update).
  - reconcile func moved into test file directly.
- Moves internal libraries to increase visibility in project.
      - pkg/api/server/internal -> pkg/internal
      - pkg/watcher/reconciler/internal -> pkg/watcher/internal
- Minor e2e test fixes found in local testing.

PipelineRun watching to be done in another PR.

Partially addresses #23


pkg/watcher/results/results_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	objs := []metav1.Object{
+		&v1beta1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "taskrun",
+				Namespace: "test",
+			},
+		},
+		&v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{
```
dibyom, line 59: Do we have a test somewhere that includes a Task/PipelineRun with a Spec/Status as well? (prob. makes more sense as an e2e test.)

dibyom (approved):
Looking good!

pkg/watcher/results/results_test.go
wlynch: Added this to the TaskRun reconciler test in pkg/watcher/reconciler/taskrun/taskrun_test.go.

pkg/watcher/internal/test/clients.go
```diff
@@ -61,7 +63,35 @@ func NewResultsClient(t *testing.T) pb.ResultsClient {
 	return pb.NewResultsClient(conn)
 }
 
-func GetFakeClients(t *testing.T, d pipelinetest.Data, client pb.ResultsClient) (context.Context, pipelinetest.Clients, *configmap.InformedWatcher) {
+func NewLegacyResultsClient(t *testing.T) v1alpha1pb.ResultsClient {
```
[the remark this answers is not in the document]
wlynch: Done.

XinruZhang (approved):
/lgtm

PR #33, opened 2020-12-29
Supports CreatedTime and UpdatedTime in Record.

XinruZhang:
Supports CreatedTime and UpdatedTime in Record.
    
This commit mainly aims to provide timestamp support when creating and updating a Record:

1. Adds CreatedTime and UpdatedTime in
  - proto definition
  - sql schema
  - gorm model definition
2. Modifies ToAPI and ToStorage to get the two timestamps work.
  - tests if the two fields can be properly transformed.
3. Updates the two fields when CreateRecord.
  - tests if the two fields are updated after the creation.
4. Updates the UpdateTime field when UpdateRecord.
  - tests if the UpdateTime field is updated after updating.

In addition, there are three minor changes:
1. To make the code structure more tidy, move TestMain from results_test.go to server_test.go
2. Solves a bug when comparing Error Code in UpdateRecord
3. Adjusts the indent in resources.proto

PR #35, opened 2021-01-06
Implements the v1alpha2 ETAG

XinruZhang:
Implements the v1alpha2 ETAG

Makes the Etag field functional to avoid race condition problem when
updating results and records, see https://google.aip.dev/154 for details.
- Adds the `ETag` in the database schema and the gorm model.
- Makes sure ToStorage and ToAPI can convert the new field properly.
- Updates etag when creating and updating a Result or Record
- The Etag provided by an Update Request has to be aligned with its
  corresponding Etag that saved in the database.

Based on #33

### DIFF
--- a/pkg/api/server/v1alpha2/results.go
+++ b/pkg/api/server/v1alpha2/results.go
@@ -57,6 +57,11 @@ func (s *Server) CreateResult(ctx context.Context, req *pb.CreateResultRequest) 
 	if err != nil {
 		return nil, err
 	}
+
+	if err := result.UpdateEtag(store); err != nil {
+		return nil, err
+	}
+
 	if err := errors.Wrap(s.db.WithContext(ctx).Create(store).Error); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


pkg/api/server/v1alpha2/results.go
```diff
@@ -90,9 +95,20 @@ func (s *Server) UpdateResult(ctx context.Context, req *pb.UpdateResultRequest)
 			return status.Errorf(codes.NotFound, "failed to find a result: %v", err)
 		}
 
+		// If the user provided the Etag field, then make sure the value of this field matches what saved in the database.
+		// See https://google.aip.dev/154 for more information.
+		if req.GetEtag() != "" {
+			if req.GetEtag() != prev.Etag {
```
wlynch, line 101:
```suggestion
		if req.GetEtag() != "" && req.GetEtag() != prev.Etag {
```

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -111,6 +113,7 @@ func TestUpdateResult(t *testing.T) {
 	tt := []struct {
 		name        string
 		requestName string // the `Nam`e field of the `UpdateResultRequest`
+		etag        string
```
wlynch, line 116: Doesn't look like this is being used in any test cases. We should add some!  (both successful and rejected)

schema/results.sql
```diff
@@ -17,10 +17,12 @@ CREATE TABLE results (
 	id varchar(64),
 
 	name varchar(64),
+	annotations BLOB,
+
 	created_time timestamp default current_timestamp not null,
 	updated_time timestamp default current_timestamp not null,
 	
-	annotations BLOB,
+	etag text,
```
wlynch, line 25: IIUC, using text limits our ability to include this in other places in the schema (notably indexes). Can we guarantee a maximum size here? If so, we might want to consider moving to varchar.

XinruZhang: Yeah, It should be 128, and done.

pkg/api/server/v1alpha2/results_test.go
XinruZhang: Done

pkg/api/server/v1alpha2/results.go
XinruZhang: Done

wlynch (approved):
/lgtm

PR #39, opened 2021-01-11
Remove v1alpha1 API.

wlynch:
We have moved over the watcher to v1alpha2, so this API is no longer
needed. Since this is alpha, no need for deprecation periods.

Cleanup from #23


XinruZhang (approved):
/lgtm

PR #43, opened 2021-01-11
Watcher: Automatically group records with known parent types.

### LINKED ISSUE #40: Group PipelineRuns with related TaskRuns
### Feature request

For TaskRuns that are created as a result of a PipelineRun, we should store them as Records within the same Result

### Use case

This aggregates related Records within the same result, which we want for querying.

wlynch:
Typically we want to group records together with other data related to
the same CI event - e.g. TaskRuns with the PipelineRun that created it,
PipelineRuns with any Trigger data that caused it to be created.

This change modifies the watcher to be aware of known annotations of
other Tekton components that would indicate a parent resource exists. If
detected, the Record will be grouped with the parent instead of creating
a new Result.

Also fixes a minor bug in the reconciler test where we were not polling
for results.

Fixes #40

### DIFF
--- a/pkg/watcher/results/results.go
+++ b/pkg/watcher/results/results.go
@@ -61,10 +61,7 @@ func (c *Client) Put(ctx context.Context, o metav1.Object, opts ...grpc.CallOpti
 // ensureResult gets the Result corresponding to the Object, or creates a new
 // one.
 func (c *Client) ensureResult(ctx context.Context, o metav1.Object, opts ...grpc.CallOption) (*pb.Result, error) {
-	name, ok := o.GetAnnotations()[annotation.Result]
-	if !ok {
-		name = result.FormatName(o.GetNamespace(), c.defaultName(o))
-	}
+	name := c.resultName(o)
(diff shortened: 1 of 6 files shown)


pkg/watcher/results/results.go
```diff
@@ -61,10 +61,7 @@ func (c *Client) Put(ctx context.Context, o metav1.Object, opts ...grpc.CallOpti
 // ensureResult gets the Result corresponding to the Object, or creates a new
 // one.
 func (c *Client) ensureResult(ctx context.Context, o metav1.Object, opts ...grpc.CallOption) (*pb.Result, error) {
-	name, ok := o.GetAnnotations()[annotation.Result]
-	if !ok {
-		name = result.FormatName(o.GetNamespace(), c.defaultName(o))
-	}
+	name := resultName(o)
```
XinruZhang, line 64: Have a question here: if a taskRun and a PipelineRun in the same namespace shares the same name, and they have nothing to do with each other, then their result name here will be the same, will this cause a problem? Like they will be put under the same result by mistake?

XinruZhang (approved):
/lgtm

PR #46, opened 2021-01-12
API: Allow uppercase letters in names.

### LINKED ISSUE #42: Result API rejects names containing upper case letters
# Expected Behavior

I should be able to use any TaskRun/PipelineRun name as a name part for a result / record identifier.

# Actual Behavior

The Result API only accepts lower case identifiers.

```
rpc error: code = InvalidArgument desc = rpc error: code = InvalidArgument desc = name must match (^[a-z0-9_-]{1,63})/results/([a-z0-9_-]{1,63}$)
```

1. Try to create any Result API resource with an upper case letter

# Additional notes

We should loosen the regex to allow upper case letters.

wlynch:
This loosens the allowed name regex to also allow for uppercase
letters, which allows us to support additional valid k8s names that
conform to RFC 1123 (see
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/).

Also adds additional tests for invalid characters in names.

Fixes #42

### DIFF
--- a/pkg/api/server/v1alpha2/record/record.go
+++ b/pkg/api/server/v1alpha2/record/record.go
@@ -32,7 +32,7 @@ import (
 
 var (
 	// NameRegex matches valid name specs for a Result.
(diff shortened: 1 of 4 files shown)


dibyom (approved):
/lgtm

Issue #41, opened 2021-01-11
Set up Watcher for Tekton Dogfooding Cluster


wlynch:
Tracking issue for setting up the Watcher on the Tekton dogfooding cluster.

Blocked on #27, #40

wlynch:
API + Watch is now running in dogfooding in shadow mode (e.g. uploading results but not updating results). I'll flip shadow mode off in a few days.

PR #47, opened 2021-01-12
Watcher: Add flag for disabling annotation updates.

wlynch:
This adds a flag to the watcher binary that allows the watcher to upload
results without annotating the original TaskRun/PipelineRun. This will
allow us to start shadowing results in the Tekton dogfooding cluster
without worrying about impacting any existing data.

/cc #41


dibyom (approved):
/lgtm

Issue #48, opened 2021-01-13
Dogfooding TaskRun names do not fit inside Result name lengths


wlynch:
# Expected Behavior

We should be able to repurpose TaskRun names / IDs as Result name parts.

# Actual Behavior

Certain names do not match the name regex because they are too large. 

Example: `tektonci/results/taskrun-check-pr-labels-rzjhs-check-kind-label-g6qxc-check-name-m-vxrhd`

Full log:
> {"level":"error","ts":"2021-01-13T10:44:08.825-0500","logger":"watcher.TaskRunResultsWatcher","caller":"taskrun/taskrun.go:62","msg":"error updating Record: rpc error: code = InvalidArgument desc = GetResult(tektonci/results/taskrun-check-pr-labels-rzjhs-check-kind-label-g6qxc-check-name-m-vxrhd): rpc error: code = InvalidArgument desc = rpc error: code = InvalidArgument desc = name must match (^[a-z0-9_-]{1,63})/results/([a-z0-9_-]{1,63}$)","knative.dev/traceid":"987d645f-0b1c-4dc3-9f1d-3c932334b4c2","knative.dev/key":"tektonci/check-pr-labels-rzjhs-check-kind-label-g6qxc-check-name-m-vxrhd","stacktrace":"github.com/tektoncd/results/pkg/watcher/reconciler/taskrun.(*Reconciler).Reconcile\n\t/usr/local/google/home/wlynch/src/github.com/tektoncd/results/pkg/watcher/reconciler/taskrun/taskrun.go:62\nknative.dev/pkg/controller.(*Impl).processNextWorkItem\n\t/usr/local/google/home/wlynch/pkg/mod/knative.dev/pkg@v0.0.0-20200831162708-14fb2347fb77/controller/controller.go:511\nknative.dev/pkg/controller.(*Impl).RunContext.func3\n\t/usr/local/google/home/wlynch/pkg/mod/knative.dev/pkg@v0.0.0-20200831162708-14fb2347fb77/controller/controller.go:449"}
```

wlynch:
The issue here is that both the Results API and k8s object names approximately follow RFC 1123, which has a max length of of 63 chars. If we want to support `<type>-<name>` syntax, we'll need to either bump the max allowed length to 127 (63+1+63) or use a different strategy for selecting default names.

wlynch:
Few options I'm considering at the moment:

[...]

PR #49, opened 2021-01-13
Revert "API: Allow uppercase letters in names."

wlynch:
This reverts commit 6ba1f3f8dbb8393f090fe8b1823e2d0a59571558.

I can't read - the [k8s name spec](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) explictly says that names are *only
lowercase*.

Other changes might need to be made in future PRs to be conformant with
RFC 952 - removal of '_', allow '.', increase allowed name part size to
255 characters. This depends on decisions made in
https://github.com/tektoncd/results/issues/48 (since we may also change
how we generate default names).

### DIFF
--- a/pkg/api/server/v1alpha2/record/record.go
+++ b/pkg/api/server/v1alpha2/record/record.go
@@ -32,7 +32,7 @@ import (
 
 var (
 	// NameRegex matches valid name specs for a Result.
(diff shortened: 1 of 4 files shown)


dibyom:
> I can't read - the k8s name spec explictly says that names are only
really? huh TIL.

dibyom (approved):
/lgtm

PR #51, opened 2021-01-13
Implements DeleteRecord

### LINKED ISSUE #9: [v1alpha2] DeleteRecord
Tracking issue for porting [Delete Record](https://github.com/tektoncd/results/blob/319047f8d1c3355d305076cbbe426c56198d9a02/proto/v1alpha2/api.proto#L68-L72) (formerly known as Executions) behavior from [v1alpha1](https://github.com/tektoncd/results/blob/319047f8d1c3355d305076cbbe426c56198d9a02/proto/v1alpha1/api.proto#L65-L68) to [v1alpha2](https://github.com/tektoncd/results/blob/319047f8d1c3355d305076cbbe426c56198d9a02/proto/v1alpha2/resources.proto#L41-L55).

XinruZhang:
This commit implements the DeleteRecord endpoint. It firstly checks
whether the record exists, if does, deletes, otherwise returns NotFound.

Closes #9

### DIFF
--- a/pkg/api/server/v1alpha2/records_test.go
+++ b/pkg/api/server/v1alpha2/records_test.go
@@ -40,7 +40,7 @@ import (
 func TestCreateRecord(t *testing.T) {
 	srv, err := New(test.NewDB(t))
 	if err != nil {
-		t.Fatalf("failed to create temp file for db: %v", err)
+		t.Fatalf("failed to create server: %v", err)
 	}
 
 	ctx := context.Background()
@@ -166,7 +166,7 @@ func TestCreateRecord_ConcurrentDelete(t *testing.T) {
 func TestGetRecord(t *testing.T) {
 	srv, err := New(test.NewDB(t))
 	if err != nil {
-		t.Fatalf("failed to create temp file for db: %v", err)
+		t.Fatalf("failed to create server: %v", err)
 	}
 
 	ctx := context.Background()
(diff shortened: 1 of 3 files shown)


pkg/api/server/v1alpha2/records_test.go
```diff
@@ -557,3 +557,48 @@ func TestUpdateRecord(t *testing.T) {
 		})
 	}
 }
+
+func TestDeleteRecord(t *testing.T) {
+	srv, err := New(test.NewDB(t))
+	if err != nil {
+		t.Fatalf("failed to create temp file for db: %v", err)
```
wlynch, line 564: nit: i'd make this error message more opaque if possible. this might have failed for other reasons besides the temp file. we use this elsewhere:

```suggestion
		t.Fatalf("failed to create server: %v", err)
```

pkg/api/server/v1alpha2/records_test.go
```diff
@@ -557,3 +557,48 @@ func TestUpdateRecord(t *testing.T) {
 		})
 	}
 }
+
+func TestDeleteRecord(t *testing.T) {
```
wlynch, line 561: A related test that would be very useful is to test whether cascade deletes work as intended (i.e. if you delete the Result, the record is deleted as well).

This isn't directly testing the DeleteRecord RPC, so up to you if you want to include it in this PR or defer to another PR.

pkg/api/server/v1alpha2/records_test.go
```diff
@@ -557,3 +557,48 @@ func TestUpdateRecord(t *testing.T) {
...
+
+	t.Run("success", func(t *testing.T) {
+		// Delete inserted record
+		if _, err := srv.DeleteRecord(ctx, &pb.DeleteRecordRequest{Name: r.GetName()}); err != nil {
+			t.Fatalf("could not delete record: %v", err)
+		}
+		// Check if the the record is deleted
+		if r, err := srv.GetRecord(ctx, &pb.GetRecordRequest{Name: r.GetName()}); err == nil {
+			t.Fatalf("expected record to be deleted, got: %+v", r)
+		}
```
wlynch, line 595: This would pass if GetRecord returned any error, including other errors besides NOT_FOUND. We probably want to do something like this instead:

```suggestion
		if r, err := srv.GetRecord(ctx, &pb.GetRecordRequest{Name: r.GetName()}); status.Code(err) != codes.NotFound {
			t.Fatalf("expected record to be deleted, got: %+v, %v", r, err)
		}
```

wlynch (commented):
Looks good! Few minor things.

pkg/api/server/v1alpha2/records_test.go
XinruZhang: I guess I can send out another PR and put the cascade delete test in the TestDeleteResult.

pkg/api/server/v1alpha2/records_test.go
```diff
@@ -557,3 +557,48 @@ func TestUpdateRecord(t *testing.T) {
...
+	}
+
+	t.Run("success", func(t *testing.T) {
+		// Delete inserted record
+		if _, err := srv.DeleteRecord(ctx, &pb.DeleteRecordRequest{Name: r.GetName()}); err != nil {
+			t.Fatalf("could not delete record: %v", err)
+		}
+		// Check if the the record is deleted
+		if r, err := srv.GetRecord(ctx, &pb.GetRecordRequest{Name: r.GetName()}); status.Code(err) != codes.NotFound {
+			t.Fatalf("expected record to be deleted, got: %+v", r)
```
wlynch, line 594: Also include the error, since this might have failed for some other reason.

wlynch (commented):
one last nit, otherwise lgtm

wlynch (approved):
/lgtm

PR #53, opened 2021-01-13
Use Object UIDs as Result/Record names.

### LINKED ISSUE #48: Dogfooding TaskRun names do not fit inside Result name lengths

wlynch:
This uses Object UIDs as the Record name in order to accomplish a few
different things:
- Fixes an issue where k8s object names (<= 253 chars) didn't fit in to Result name
  parts (<= 63 chars). UIDs
- Sidesteps the original reason why we tried prepending the type in the
  first place to distinguish TaskRuns/PipelineRuns with the same name.
  (UIDs are unique for the cluster).
- Allows us to detect and distinguish cases where users delete then
  recreate objects with the same name. Using UIDs will cause us to
  create new Results/Records for each new instance, instead of trying to
  update the existing data.

The downside of this change means that Result/Record names are no longer
human readable, which means that users will need to rely more on
Record data to distinguish entries.

Fixes #48


docs/watcher/README.md
```diff
@@ -11,10 +11,10 @@ The Watcher currently supports the following types:
...
 
-The Watcher uses annotations to automatically detect and group related Records into the same Result. The following annotations are recognized (in order of precedence):
+The Watcher uses Object data to automatically detect and group related Records into the same Result. The following data is checked (in order of precedence):
 
-- `results.tekton.dev/result`
-- `triggers.tekton.dev/triggers-eventid`
-- `tekton.dev/pipelineRun`
+- `results.tekton.dev/result` annotation. This should correspond to the full `Result.name` identifier (e.g. `foo/results/bar`).
+- `triggers.tekton.dev/triggers-eventid` annotation (this is generated from Objects created via [Tekton Triggers](https://github.com/tektoncd/triggers))
+- An [OwnerReference](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents) to a PipelineRun.
```
dibyom, line 18: minor: think TaskRuns created by a PipelineRun will also have a `tekton.dev/pipelineRun` label added that could be used instead (might make this a tad easier since we are already using labels/annotations for the other two).

docs/watcher/README.md
```diff
@@ -11,10 +11,10 @@ The Watcher currently supports the following types:
...
 ## Result Grouping
 
-The Watcher uses annotations to automatically detect and group related Records into the same Result. The following annotations are recognized (in order of precedence):
+The Watcher uses Object data to automatically detect and group related Records into the same Result. The following data is checked (in order of precedence):
 
-- `results.tekton.dev/result`
-- `triggers.tekton.dev/triggers-eventid`
-- `tekton.dev/pipelineRun`
+- `results.tekton.dev/result` annotation. This should correspond to the full `Result.name` identifier (e.g. `foo/results/bar`).
+- `triggers.tekton.dev/triggers-eventid` annotation (this is generated from Objects created via [Tekton Triggers](https://github.com/tektoncd/triggers))
```
dibyom, line 17: Triggers adds this as a label not an annotation.

dibyom (commented):
I think there is one issue around using labels vs annotations that we need to fix but otherwise LGTM.

docs/watcher/README.md
wlynch: Yeah I thought about doing this, but decided against it because using OwnerReference gives us direct access to the PipelineRun UID, avoiding a additional API call and any race conflicts that might arise from that.

docs/watcher/README.md
wlynch: Good catch! Fixed.

dibyom (approved):
/lgtm

Issue #27, opened 2020-12-22
Setup Release Infrastructure


wlynch:
We need to set up release infrastructure to publish API Server and Watcher images to the Tekton repository.

We can likely use Pipelines/Triggers/Plumbing as a model to follow.

wlynch:
Closing this out since we have enough pieces in place to do a v0.1.x release. Additional release automation will be tracked in https://github.com/tektoncd/results/issues/70

PR #55, opened 2021-01-19
Add Tekton CI/Release configs.

wlynch:
This adds initial configs for building and deploying Docker images for
the API Server and Watcher.

Verified this works on a personal registry. TODO for figuring out where
in tekton-releases this should be pointed at.

This pipeline depends on GKE Workload Identity for automatically
retrieving credentials for pushing the image. Additional work might be
needed to make this GKE-agnostic (but since we currently use GKE/GCR for
Tekton dogfooding, this should be okay for now).

Partially addresses #27

### DIFF
--- a/tekton/release-api.yaml
+++ b/tekton/release-api.yaml
@@ -1,0 +1,39 @@
+# Copyright 2021 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: results-release-api-
+spec:
+  serviceAccountName: tekton-releaser
+  pipelineRef:
+    name: results-release
+  params:
+    - name: revision
+      value: main
+    - name: package
+      value: github.com/tektoncd/results/cmd/api
+    - name: docker_repo
+      value: ko.local
+  workspaces:
+    - name: ws
+      volumeClaimTemplate:
+        spec:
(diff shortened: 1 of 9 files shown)


tekton/release-api.yaml
```diff
@@ -0,0 +1,39 @@
...
+  serviceAccountName: tekton-releaser
+  pipelineRef:
+    name: results-release
+  params:
+    - name: revision
+      value: main
+    - name: package
+      value: github.com/tektoncd/results/cmd/api
+    - name: docker_repo
+      value: <TODO>
```
dibyom, line 30: was this intentional? Should it be gcr.io/tekton-releases or gcr.io/tekton-releases-nightly?

dibyom: or ko.local?

tekton/release-watcher.yaml
```diff
@@ -0,0 +1,39 @@
...
+  serviceAccountName: tekton-releaser
+  pipelineRef:
+    name: results-release
+  params:
+    - name: revision
+      value: main
+    - name: package
+      value: github.com/tektoncd/results/cmd/watcher
+    - name: docker_repo
+      value: us-docker.pkg.dev/wlynch-test/tektoncd-results
```
dibyom, line 30: ko.local perhaps?

tekton/vendor/git-clone.yaml
```diff
@@ -0,0 +1,144 @@
+# https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2
```
dibyom, line 1: instead of vendoring the files, we could add instructions in the Readme (or a script) to run kubectl commands `kubectl apply -f  https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2 (or SHA)` 

That should make it less likely to accidentally override the tasks already applied to the dogfooding cluster.

dibyom (commented):
just a couple of comments on the docker_repo values

tekton/release-api.yaml
wlynch: Done! (this will eventually be one of the tekton-releases repos, but for now ko.local works)

tekton/release-watcher.yaml
wlynch: Done.

tekton/vendor/git-clone.yaml
wlynch: Good idea! Done.

PR #56, opened 2021-01-20
API: Allow users to list records across results.

### LINKED ISSUE #50: ListRecords across Results
### Feature request

We should add support for reading records across namespaces, so that we can query for records across the database. 

e.g. `GET /default/results/-/records?filter=...`

See https://google.aip.dev/159 for more details.

### Use case

This allows users to make queries for a particular type of record. e.g. list all taskruns with a particular property. Currently, users must call ListRecords on each result individually and aggregate themselves.

wlynch:
Follows the guidance of https://google.aip.dev/159 to allow ListResults
to read across results by specifying `-` as the Result name.

Fixes #50


pkg/api/server/v1alpha2/records_test.go
```diff
@@ -602,3 +603,53 @@ func TestDeleteRecord(t *testing.T) {
 		}
 	})
 }
+
+// TestListRecords_multiresult tests listing records across multiple result parents.
```
XinruZhang, line 607: Since we don't support List cross parents, "across multiple result parents" might be a kind of misleading?

XinruZhang (commented):
Mostly LGTM!

pkg/api/server/v1alpha2/records_test.go
wlynch: Reworded. We're listing across multiple record parents (i.e. results), but not over multiple result parents.

PR #57, opened 2021-01-20
Add health checking endpoints to RPC server.

### LINKED ISSUE #45: API Server Health Checking
### Feature request

We should add [gRPC health checking](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) to the API server for interop with other gRPC tooling.

### Use case

This would let us take advantage of other gRPC health check tooling (https://github.com/grpc-ecosystem/grpc-health-probe, grpc_cli, etc.)

wlynch:
See https://github.com/grpc/grpc/blob/master/doc/health-checking.md for
gRPC health checking documentataton.

Fixes #45

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -27,6 +27,8 @@ import (
 	v1alpha2pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/reflection"
 	"gorm.io/driver/mysql"
 	"gorm.io/gorm"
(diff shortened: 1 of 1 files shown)


ImJasonH:
/lgtm

Will/can this healthcheck endpoint also be used as a k8s healthcheck to know when to route requests through the service?

wlynch:
> Will/can this healthcheck endpoint also be used as a k8s healthcheck to know when to route requests through the service?

Yes it can! See https://github.com/grpc-ecosystem/grpc-health-probe

ImJasonH:
> > Will/can this healthcheck endpoint also be used as a k8s healthcheck to know when to route requests through the service?
> 
> Yes it can! See https://github.com/grpc-ecosystem/grpc-health-probe

Oh that's kind of gross, it requires the `grpc_health_probe` binary to be included in the image and called with an `exec` probe. 😒 

Since it's not that simple I won't push too hard for it 😄 

Still /lgtm

Issue #58, opened 2021-01-21
API: Authentication and Authorization


wlynch:
### Feature request

This issue tracks adding authn/z to the Results API.

Admins should be able to control fine grain ACLs for reading or writing data to the Result API.

### Use case

Allow the Results API to be exposed outside of the cluster so that off cluster clients may access data.

PR #60, opened 2021-01-22
API: Add support for RBAC Auth.

wlynch:
This uses Kubernetes
[TokenReview](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication)
and
[SubjectAccessReview](https://kubernetes.io/docs/reference/access-authn-authz/authorization/)
APIs to authenticate and authorize incoming user tokens using built-in
cluster RBAC.

While this change adds and tests the support, it does not enable it since
additional changes are needed to watcher to support the new behavior.
User documentation will be added when this behavior is added.

Additional changes:
- Moves pkg/watcher/internal/test to pkg/internal/test in order to make
this visible to API integration tests (necessary since gRPC metadata
only works if we are using a real gRPC client).
- Fixes issue with Delete methods erroring when returning on nil Empty
responses (caught because of the new gRPC client based tests).
- Creates new Service Accounts and Roles for the API Server and Watcher
(previously we were reusing the Pipeline Controller SA).

Partially addresses #58


wlynch:
/hold

Service account separate turned out to be a little more involved - split out in https://github.com/tektoncd/results/pull/61

(this is still ready for review for the auth impl though)

config/100-api-serviceaccount.yaml
```diff
@@ -19,6 +19,20 @@ metadata:
   namespace: tekton-pipelines
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: tekton-results-api
+rules:
+  # Results API needs to be able to verify incoming auth tokens.
+  - apiGroups: ["authentication.k8s.io"]
+    resources: ["tokenreviews"]
```
dibyom, line 28: question: do we need these? My understanding is that tokenreviews and subjectaccessreviews are just request/response formats (like the interceptorrequest format) while the `rules` only apply to actual resources that the API server or the RBAC system knows about. (I don't actually know a ton about this so I could be totally off base here).

config/100-watcher-serviceaccount.yaml
```diff
@@ -23,6 +23,10 @@ kind: ClusterRole
 metadata:
   name: tekton-results-watcher
 rules:
+  # Watcher needs to be able to create new and update existing results.
+  - apiGroups: ["results.tekton.dev"]
```
dibyom, line 27: Interesting, so it seems like the resources don't have to map to actual K8s CRD resources 🤯

config/100-api-serviceaccount.yaml
dibyom: Ok, so so looks like rules can apply to non CRD types too as in the file below. But I guess I still have the question: do we need these two roles to make the SAR /TokenReview calls?

dibyom (commented):
Impl looks solid. Overall the idea of piggy backing off of k8s RBAC is really interesting. It does introduce a bit of an inconsistency i.e. you now have results.tekton.dev which is not a CRD but part of your service account roles. Not sure how prevalent this pattern is but for our use case this seems like a pretty good fit.

dibyom (approved):

config/100-api-serviceaccount.yaml
wlynch: The request/response format would come into play if we were writing an admission control webhook to handle the auth. In this case, this is happening directly in the API Server.

TokenReviews only expose "`create`" (POST) method as the mechanism for issuing review requests. This RBAC Role controls who is allowed to make these calls (otherwise you get an error like `tokenreviews.authentication.k8s.io is forbidden: User "system:serviceaccount:tekton-pipelines:tekton-results-api" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope`). SAR reviews are similar.

config/100-watcher-serviceaccount.yaml
wlynch: Yup! I was really please this Just Worked™️

config/100-api-serviceaccount.yaml
dibyom: ahh got it! Interesting!

PR #61, opened 2021-01-26
Separate result configs from pipelines.

wlynch:
This modifies the config yamls to do a few things:
- Use result specific Service Accounts instead of relying on
tekton-pipeline-controll.
- Creates custom [Cluster]Roles/Bindings to grant minimum permissions.
- Updates labels to use `results.tekton.dev` prefix.


XinruZhang (approved):
/lgtm

PR #63, opened 2021-01-26
Add installation docs, move mysql-init to config/.

wlynch:
This adds the initial placeholder documentation for installation
instructions from release sources (as opposed to only installation from
source).

Additionally, this moves the mysql init step to its own config file,
automatically generated from the sql config. I plan to add a test for
all autogenerated config files in another PR, so this does not include
any test at the moment (beyond the e2e test).


dibyom (approved):
/lgtm

PR #69, opened 2021-01-27
Move top level README to docs.

wlynch:
This is necessary for the Tekton website to render this page properly
(it assumes everything is rooted in 1 folder). See https://github.com/tektoncd/website/pull/230

GitHub supports rendering project READMEs stored in the docs folder
(https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-readmes)
so this should have minimal impact on the browser UX.


dibyom (approved):
/lgtm

interesting: AFAIK for pipelines/triggers development.md is not in docs/ we should maybe do that.

PR #72, opened 2021-02-01
Add simple UI for visualizing Result data.

wlynch:
This folder contains a test HTTP server for rendering Results in a very basic
UI.

This is **not** intended to be a replacement UI for
[dashboard](https://github.com/tektoncd/dashboard) - this is only here for
local testing to help visualize result data.

This is only intended for testing, so testing rigor and polish are not
strictly enforced.

### DIFF
--- a/tools/simpleui/main.go
+++ b/tools/simpleui/main.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"fmt"
+	"html/template"
+	"log"
+	"net/http"
+	"strings"
+
+	"github.com/gorilla/mux"
+	pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
+	"google.golang.org/grpc"
+)
+
+var (
+	resultTmpl = template.Must(template.New("results.html").ParseFiles("templates/results.html"))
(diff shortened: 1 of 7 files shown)


tools/simpleui/main.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	r := mux.NewRouter()
+	r.HandleFunc("/", u.home)
+	r.HandleFunc("/{namespace}", u.results)
+	r.HandleFunc("/{namespace}/results/{name}", u.records)
+	http.Handle("/", r)
+
+	hostname, _ := os.Hostname()
+	log.Printf("Running on http://%s:8080\n", hostname)
```
XinruZhang, line 78: Tiny error here: the hostname doesn't work for me: 

> 2021/02/01 16:12:11 Running on http://xinruzhang-macbookpro.roam.corp.google.com:8080

Using `localhost` instead solved the problem. Is there any other function can fit this requirement?

XinruZhang (commented):
Looks great!

tools/simpleui/main.go
wlynch: Done.

PR #78, opened 2021-02-02
Enable TLS and use Service Account tokens in Watcher.

### LINKED ISSUE #58: API: Authentication and Authorization

wlynch:
- Configures watcher to use the Pod Service Account token
(https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod)
to authenticate to the API Server.
- Changes the authorization header format API Server expects to be
inline with gRPC TokenSource (OAuth2 Bearer).
- Since gRPC forbids usage of TokenSource transport without SSL enabled,
configures usage of TLS in gRPC client/server (see
https://grpc.io/docs/guides/auth/#with-server-authentication-ssltls).
  - By default, docs guide users to generate a self-signed key/cert
  pair.
- Includes premade ClusterRoles for common auth roles we expect (ro, rw,
admin).

Fixes #58

### DIFF
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -18,10 +18,12 @@ package main
 
 import (
 	"context"
-	"crypto/tls"
 	"crypto/x509"
 	"flag"
+	"fmt"
+	"io/ioutil"
 	"log"
+	"os"
 	"time"
 
 	creds "github.com/tektoncd/results/pkg/watcher/grpc"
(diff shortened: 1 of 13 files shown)


cmd/watcher/main.go
```diff
@@ -105,3 +122,26 @@ func connectToAPIServer(ctx context.Context, apiAddr string, authMode string) (*
...
 }
+
+func loadCerts() (*x509.CertPool, error) {
+	// Setup TLS certs to the server.
+	f, err := os.Open("/etc/tls/tls.crt")
+	if err != nil {
+		log.Println("no local cluster cert found, defaulting to system pool...")
+		return x509.SystemCertPool()
+	}
+	defer f.Close()
```
dibyom, line 133: nit: shouldn't this defer be before the err != nil line so that if we hit the err block, we still call f.Close before returning?

dibyom (approved):

cmd/watcher/main.go
wlynch: Doesn't make a difference. If err is returned then [f will be nil](https://go.googlesource.com/go/+/go1.15.7/src/os/file.go#329), so the [close won't do anything and return an error](https://go.googlesource.com/go/+/go1.15.7/src/os/file_posix.go#21).

Issue #85, opened 2021-02-05
[api] List ordering


wlynch:
A common use case will be to order results by creation time (e.g. show me the most recent).

We should look into adding an ordering field to List methods to support this - https://google.aip.dev/132#ordering

sbwsg:
I'll take a look into implementing this!

/assign

wlynch:
Thanks Scott! Places we'll likely need changes: 

- Proto definition for [ListResults](https://github.com/tektoncd/results/blob/da219e11cb33f1ecb4a6815e3b39ad77b417d667/proto/v1alpha2/api.proto#L120) and [ListRecords](https://github.com/tektoncd/results/blob/da219e11cb33f1ecb4a6815e3b39ad77b417d667/proto/v1alpha2/api.proto#L162) (also [regen the proto libraries](https://github.com/tektoncd/results/blob/main/proto/README.md#regenerating-proto-libraries))
- DB queries made for [Results](https://github.com/tektoncd/results/blob/da219e11cb33f1ecb4a6815e3b39ad77b417d667/pkg/api/server/v1alpha2/results.go#L216) and [Records](https://github.com/tektoncd/results/blob/da219e11cb33f1ecb4a6815e3b39ad77b417d667/pkg/api/server/v1alpha2/records.go#L192)

PR #87, opened 2021-02-09
Add order_by to results and records

sbwsg:
Prior to this commit results and records could not be requested with an ordering.

This commit introduces an order_by clause that accepts a field name and an optional (asc|desc). This follows the AIP describing ordering here: https://google.aip.dev/132#ordering

Fixes https://github.com/tektoncd/results/issues/85


sbwsg:
Hm, it looks like protoc is a different version.  Let me see if I can figure out how to generate this with a matching version to the original file.

sbwsg:
Looks like the markdown link checker experienced some network weather while checking sqlite.com.

/test pull-tekton-results-build-tests

pkg/api/server/v1alpha2/ordering.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func orderBy(fields string) (string, error) {
```
wlynch, line 25: Let's add tests for these to be thorough.

pkg/api/server/v1alpha2/ordering.go
```diff
@@ -0,0 +1,61 @@
...
+		return "", nil
+	}
+
+	var orderBy []string
+	for _, field := range strings.Split(fields, ",") {
+		field = strings.TrimSpace(field)
+		f := strings.Split(field, " ")
+		switch len(f) {
+		case 1:
+			orderBy = append(orderBy, strings.TrimSpace(f[0]))
```
wlynch, line 36: I think we're going to need a bit more logic here, since the database schema doesn't match the API schema:

- the `name` field in the API for results actually maps to (`parent`, `name`) in the database schema, so names can potentially be out of order depending on their parent. 
- Timestamps have the reverse problem- the API timestamp has the subfields (`seconds`, `nanos`), but the timestamp in the schema is closer to a `time.Time`.

We'll also want some documentation and validation to give clear errors on what fields are supported - I don't think we have the ability to order by fields in the data blob.

wlynch (commented):
Great start - looking good so far! :D

pkg/api/server/v1alpha2/ordering.go
sbwsg: Ah ok, thanks for this.  I need to dig around and understand what's going on a bit more thoroughly.

sbwsg:
I realise now that I completely missed that the original issue specifies the order_by should only be available for creation time.

wlynch:
> I realise now that I completely missed that the original issue specifies the order_by should only be available for creation time. d'oh.

I think it's great to support more fields! creation time is just primary motivation. I think it'd be fine to support any top level fields in Result/Record (e.g. anything now embedded in `data`). As long as we document which fields are supported, any subset of that is fine with me!

sbwsg:
Sorry for the delay working on this.  Somehow I've gotten myself in to a situation where I cannot get `watcher` to connect to `api` _at all_.  I'm having this problem in two separate kind clusters running on two different machines (one mac and one linux) and in neither case will the `watcher` stay alive more than 30 seconds before hitting a context deadline and exiting:

```bash
2021/03/02 14:32:55 dialing tekton-results-api-service.tekton-pipelines.svc.clu
ster.local:50051...
2021/03/02 14:33:25 did not connect: context deadline exceeded
```

And the errors I get from the grpc_cli when trying to communicate with the `api`:

```bash
λ grpc_cli call --channel_creds_type=ssl --ssl_target=tekton-results-api-service.tekton-pipelines.svc.cluster.local --call_creds=access_token=$(kubectl get secrets -n tekton-pipelines -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='tekton-results-watcher')].data.token}"|base64 --decode) localhost:50051 tekton.results.v1alpha2.Results.ListResults 'parent: "default"'
connecting to localhost:50051
Rpc failed with status code 16, error message: permission denied
```

pkg/api/server/v1alpha2/ordering.go
sbwsg: I've limited the fields which can be ordered to just `created_time` and `updated_time` now.  So this limits the field types we're sorting on to just timestamps in this PR.

I'm curious about the API timestamp vs schema timestamp issue you mention - I'm not totally clear where that comes in to play yet.

I've added a short documentation snippet to `docs/api/README.md` describing the order_by parameter.

pkg/api/server/v1alpha2/ordering.go
sbwsg: Tests for the functions in this file have now been added.

pkg/api/server/v1alpha2/ordering.go
```diff
@@ -0,0 +1,99 @@
...
+		errmsg := fmt.Sprintf("order by %s not supported", fieldName)
+		return "", status.Error(codes.InvalidArgument, errmsg)
+	}
+
+	if direction == "" {
+		return fieldName, nil
+	} else {
+		direction = strings.TrimSpace(direction)
+		return orderByDirection(fieldName, direction)
+	}
```
wlynch, line 77: nit:
```suggestion
	if direction == "" {
		return fieldName, nil
	}
	direction = strings.TrimSpace(direction)
	return orderByDirection(fieldName, direction)
```

pkg/api/server/v1alpha2/ordering.go
```diff
@@ -0,0 +1,99 @@
...
+		fieldName = f[0]
+	case 2:
+		fieldName = f[0]
+		direction = f[1]
+	default:
+		errmsg := fmt.Sprintf("invalid order_by %q", field)
+		return "", status.Error(codes.InvalidArgument, errmsg)
+	}
+
+	fieldName = strings.TrimSpace(fieldName)
```
wlynch, line 65: IIUC this isn't necessary, since `strings.Fields` will already trim any consecutive whitespace.

pkg/api/server/v1alpha2/ordering.go
```diff
@@ -0,0 +1,99 @@
...
+	for i := range allowedOrderByFields {
+		if name == allowedOrderByFields[i] {
+			return true
+		}
+	}
+	return false
+}
+
+func orderByDirection(field string, direction string) (string, error) {
+	sortOrder := strings.ToLower(strings.TrimSpace(direction))
```
wlynch, line 90: The `TrimSpace`s in this func is redundant with the trimming already done in `normalizeOrderByField`. Can we consolidate these?

pkg/api/server/v1alpha2/ordering.go
```diff
@@ -0,0 +1,99 @@
...
+	direction := ""
+	switch len(f) {
+	case 1:
+		fieldName = f[0]
+	case 2:
+		fieldName = f[0]
+		direction = f[1]
+	default:
+		errmsg := fmt.Sprintf("invalid order_by %q", field)
+		return "", status.Error(codes.InvalidArgument, errmsg)
```
wlynch, line 62:
```suggestion
		return "", status.Errorf(codes.InvalidArgument, "invalid order_by %q", field)
```

pkg/api/server/v1alpha2/ordering.go
```diff
@@ -0,0 +1,99 @@
...
+
+func orderByDirection(field string, direction string) (string, error) {
+	sortOrder := strings.ToLower(strings.TrimSpace(direction))
+	switch sortOrder {
+	case "asc":
+		return fmt.Sprintf("%s ASC", strings.TrimSpace(field)), nil
+	case "desc":
+		return fmt.Sprintf("%s DESC", strings.TrimSpace(field)), nil
+	default:
+		return "", status.Error(codes.InvalidArgument, fmt.Sprintf("invalid sort direction %q", sortOrder))
```
wlynch, line 97:
```suggestion
		return "", status.Errorf(codes.InvalidArgument, "invalid sort direction %q", sortOrder)
```

pkg/api/server/v1alpha2/results_test.go
```diff
@@ -325,7 +325,9 @@ func TestListResults(t *testing.T) {
 
 	parent := "foo"
 	results := make([]*pb.Result, 0, 5)
+
 	for i := 1; i <= cap(results); i++ {
+		fakeClock.Advance(1)
```
wlynch, line 330: super nit: Prefer using a qualified unit so it's clear how much we're incrementing (e.g. `1*time.Second`). 

Since Duration's are nanoseconds, incrementing by such a small number may cause problems if we're not careful about preserving the precision (e.g. if unix time is ever introduced anywhere).

wlynch (commented):
Looks great! 🎉 
Just some minor changes.

pkg/api/server/v1alpha2/ordering.go
sbwsg: Ach, good point, removed all the no-op `strings.TrimSpace`

pkg/api/server/v1alpha2/ordering.go
sbwsg: Thanks for this, I did not realise there was a `status.Errorf`

pkg/api/server/v1alpha2/results_test.go
sbwsg: Good point, I prefer the explicitness of `time.Second` over `1`.

sbwsg:
@wlynch I think all points are now addressed.  Lmk if there's anything else you'd want done here.

pkg/api/server/v1alpha2/records_test.go
```diff
@@ -255,6 +255,7 @@ func TestListRecords(t *testing.T) {
 	records := make([]*pb.Record, 0, 6)
 	// Create 3 TaskRun records
 	for i := 0; i < 3; i++ {
+		fakeClock.Advance(1)
```
wlynch, line 258: Few other instances of `Advance(1)` that we might want to change to `Advance(time.Second)`

wlynch (commented):
LGTM, just one tiny nit.

/lgtm

pkg/api/server/v1alpha2/records_test.go
sbwsg: D'oh, good catch!  All uses of `Advance(1)` should be replaced now.

wlynch (approved):
/lgtm

PR #88, opened 2021-02-09
Add background to datamodel image

sbwsg:
Prior to this commit the datamodel png had a transparent background
which meant the diagram couldn't be easily read when Github is in
dark mode.

This commit updates the image to have a white background so it's
legible when dark mode is enabled.  Generated with `convert ./docs/images/datamodel.png -background "#ffffff" -flatten ./docs/images/datamodel-white.png`

Screenshot 2021-02-09 1 34 12 PM


wlynch (approved):
/lgtm

PR #92, opened 2021-02-22
e2e tests: Rewrite e2e tests in Go.

wlynch:
This change rewrites the e2e tests to be written in Go so that we can
have better typing and (hopefully) make it easier to add logic and
conditional behavior in the future.

For now this uses the typed client, we might look into using a dynamic
client to generalize this a bit more in the future.


ImJasonH (commented):
/lgtm

Issue #59, opened 2021-01-21
Watcher: Pipeline etcd cleanup


wlynch:
### Feature request

After a TaskRun/PipelineRun is completed and successfully uploaded, we should delete the etcd based data to free up resources on the user's cluster.

Requirements:
- etcd based data should not be deleted until the Task/PipelineRun is complete and we've confirmed that the data has been successfully uploaded to the API.
- There should be a configurable grace period for how long we wait before we deleting so that other integrations relying on the CRD API directly don't break.

### Use case

Task/PipelineRun clean up has been a much requested feature, but is at odds with results storage since once the CRD object is gone from etcd, the results are gone with it.

Making this a feature of the Results watcher means that we can ensure that this clean up happens after results have been stored durably. 

See:
- https://github.com/tektoncd/pipeline/issues/2332
- https://github.com/tektoncd/pipeline/pull/1486
- https://github.com/tektoncd/pipeline/issues/2332
- https://github.com/tektoncd/pipeline/pull/1429
- https://github.com/tektoncd/experimental/issues/479
- https://github.com/tektoncd/pipeline/issues/2452

PR #93, opened 2021-02-22
Add kustomize support for e2e tests.

wlynch:
For e2e tests, we want to match the upstream deployment as close as
possible. However, there will often be times where we need to make
minor modifications to the e2e tests to enable/modify settings.

The main motivator for this are upcoming e2e tests for automatic Run
cleanup (#59). We want to be able to significantly reduce the grace period for
Run deletion (e.g. O(seconds)) so that we can speed up test execution.
This kind of setting is probably not acceptable for typical users, who
would likely want O(minutes) if not O(hours).

This currently adds a flag that is effectively a no-op, since it is the
default value of the flag anyway.

### DIFF
--- a/config/default-clusterroles.yaml
+++ b/config/default-clusterroles.yaml
@@ -1,3 +1,17 @@
+# Copyright 2021 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


dibyom (commented):
Neat!

Do we have to do something to not include the kustomization file from the results release? Or should that already not be included by default?

/approve

wlynch:
> Do we have to do something to not include the kustomization file from the results release? Or should that already not be included by default?

`ko` is aware of kustomize and respects it in commands (incl resolve and apply).

PR #94, opened 2021-02-23
simpleui: Add query, parent fields.

wlynch:
Adds fields for linking to Record parents and query field for filtering
results.

This is what I've been using for demos. Committing so it's upstream for other people to try out!

### DIFF
--- a/tools/simpleui/template.go
+++ b/tools/simpleui/template.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 
 	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/types/known/anypb"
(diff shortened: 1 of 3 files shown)


tools/simpleui/template.go
```diff
@@ -17,3 +18,12 @@ func textproto(a *anypb.Any) (string, error) {
 	}
 	return prototext.Format(m), nil
 }
+
+func parent(in string) string {
+	s := strings.Split(in, "/")
+	return strings.Join(s[:3], "/")
```
ImJasonH, line 24: Will `in` always have 4 elements?

```
func parent(in string) string {
	return in[:strings.LastIndex(in, "/")]
}
```

wlynch: Done!

wlynch: Ended up reverting this. We need the first 3 parts, not "everything but the last part" (there are 5 parts totally for this ID).

I think we can make a reasonable assumption about this format since it's validated at write time.

XinruZhang (approved):

PR #95, opened 2021-03-02
Add Subject Alternative Name to test certificate script

sbwsg:
Prior to this commit we generated a test certificate with
openssl using the -subj flag. Depending on the go version
and environment variables this could cause issues because
the CN certificate field is deprecated in favor of the SAN field.

This commit adds the -addext flag to additionally include a
Subject Alternative Names field to the cert.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -38,7 +38,8 @@
    -out cert.pem \
    -days 365 \
    -nodes \
-   -subj "/CN=tekton-results-api-service.tekton-pipelines.svc.cluster.local"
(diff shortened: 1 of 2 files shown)


wlynch:
We should also update the installation instructions as well - https://github.com/tektoncd/results/blob/main/docs/install.md

sbwsg:
> We should also update the installation instructions as well - https://github.com/tektoncd/results/blob/main/docs/install.md

👍 Done

wlynch (approved):
/lgtm

PR #97, opened 2021-03-04
e2e: Add TaskRun for running kind tests.

wlynch:
This adds a TaskRun for executing kind tests. To start, this is a single
TaskRun, but I plan to break this out into its own reusable Task for the
catalog in the future.

The configs here are heavily based on the configs in
https://github.com/kind-ci/examples and
https://github.com/kubernetes-sigs/kind/issues/303.

This was done in favor of the Prow based config, since the cluster
failed to come up, dispite the various labels set (likely the settings
weren't applying / we are missing some other config in our prow
cluster). Since we want to move towards Tekton based CI in the future
anyway, it was simpler to handle container setup in a Task from scratch.

Much of the component installation is unfortunately copied verbatim
from the install script - there is some strange interaction going on
between kind and ko that causes ko to fail if called or sourced
from e2e.sh. I plan to investigate this further in another PR.


test/e2e/e2e.sh
```diff
@@ -0,0 +1,85 @@
...
+main() {
+    ROOT="$(git rev-parse --show-toplevel)"
+
+    # Setup
+    kind create cluster --loglevel=debug --config "${ROOT}/test/e2e/kind-cluster.yaml"
+    kind export kubeconfig
+
+    # Install
+
+    # TODO: ko fails if we try to run/source install.sh, but doesn't if we inline verbatim.
```
sbwsg, line 25: I wonder if this is because the install script now sets KIND_CLUSTER_NAME to "tekton-results" if it hasn't been set elsewhere.  ( I made this change in my most recent PR to the script because otherwise running 00-setup.sh followed by 01-install.sh resulted in an error locally because I didn't have KIND_CLUSTER_NAME defined or exported ).

It looks like the setup step here is creating a cluster named the default, "kind"?  Wonder if a simple `export KIND_CLUSTER_NAME="kind"` at the top of this script would resolve the issue running `install.sh`?

wlynch: 🤦  that totally did it!

test/e2e/tekton/taskrun.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - image: docker.io/alpine/git
+        name: checkout
+        script: |
+          git clone $(params.repo) .
+          git checkout $(params.ref)
+      # TODO: Move image to tekton-dogfooding (or other similar location)
+      - image: gcr.io/wlynch-test/kind-runner
+        name: kind
+        securityContext:
+          privileged: true
```
vdemeester, line 42: Is it required for `kind` itself ? we may need to run as root, but I don't think we need the extra capabilities `privileged` adds, maybe ?

wlynch: Looks like we can remove this! 🎉

vdemeester (approved):
/meow

vdemeester (approved):

XinruZhang (approved):

PR #98, opened 2021-03-10
Document applying all with kustomize and use `kubectl kustomize` instead

sbwsg:
I forgot we added the kustomization and started seeing
the following error when `ko apply`ing the `./config` directory:

```
error: error validating "STDIN": error validating data: [apiVersion not
set, kind not set]; if you choose to ignore these errors, turn validation
off with --validate=false
```

This happens because the kustomization.yaml in the config directory isn't
formatted like a kubernetes resource and `ko` tries reading it like one when
applying all the files in that dir.  Added a note to the development doc to
capture applying all.

Also didn't have `kustomize` installed but it turns out `kubectl` comes with
it since 0.14 (current stable version is 0.20.4), so I updated the e2e script
to use `kubectl`'s built-in version.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -36,12 +36,22 @@ for configurable options for these scripts.
 
 ### Deploying individual components
 
-You can redeploy individual components via ko. Just make sure to use
+You can redeploy individual components via ko. Just make sure to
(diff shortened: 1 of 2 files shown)


wlynch (approved):
/lgtm

Issue #81, opened 2021-02-04
[integrations] Tekton CLI


wlynch:
Tracking issue for integrating with tkn, the [Tekton CLI](https://github.com/tektoncd/cli).

We'd like to add support for querying the Results API via tkn. This can likely be done in a non-invasive way via [tkn plugins](https://github.com/tektoncd/cli/commit/247384276c1d4cbf4855390bc20e765c6881fe66).

Interesting things to figure out here include:
- How can users connect to the API if it's local to the cluster? Does it need to be a publicly available service, or is there a lightweight way to proxy commands 
- Will authentication work out of the box? Do we need to add client configuration for what service account to use, or can we make server side changes to support more types of credentials?

nikhil-thomas:
@piyush-garg, @vinamra28

wlynch:
https://github.com/tektoncd/results/tree/main/tools/tkn-results

PR #100, opened 2021-03-18
tkn-results: Add CLI for listing results and records.

wlynch:
This adds a new tkn plugin for interacting with the Results API.
For now this only implements the list methods for results and records.

I wasn't sure what the best way to go about testing the whole CLI, since
most of the complexity is setting up the client. We might want to
consider adding this as a part of the e2e tests in a future PR.

Implements #81


ImJasonH:
Is the intention for this to be a CLI for *testing/playing around with* the API? If so it probably doesn't need super thorough tests, and should just be documented as test-only.

If it's not, we probably need to have a longer discussion with the tkn folks about what the relationship is between the two.

tools/tkn-results/README.md
```diff
@@ -0,0 +1,13 @@
+# tkn-results
+
+tkn CLI plugin for interacting with the Tekton Results API.
+
+## Installation
+
+```
+$ GOBIN=${TKN_PLUGINS_DIR:-"${HOME}/.config/tkn/plugins"} go install github.com/tektoncd/experimental/tkn-resolve
```
afrittoli, line 8: Is `github.com/tektoncd/experimental/tkn-resolve` the right path?

wlynch: Good catch! 😅  Fixed.

wlynch:
> Is the intention for this to be a CLI for _testing/playing around with_ the API? If so it probably doesn't need super thorough tests, and should just be documented as test-only.
> 
> If it's not, we probably need to have a longer discussion with the tkn folks about what the relationship is between the two.

Short term this is to let users have a nicer mechanism to work with the Results API without needing to jump through grpc_cli hoops. There's a blurry line here between a test-only tool and the recommended way to see results on a CLI.

I agree that we should have a longer discussion about how results fits in to the existing tkn UX for pipelineruns/taskruns (particularly since results is an optional component), but for now our hope is to start prototyping with a plugin.

tools/tkn-results/cmd/records_list.go
```diff
@@ -0,0 +1,42 @@
...
+				fmt.Printf("ListRecords: %v\n", err)
+				return err
+			}
+			return printproto(os.Stdout, resp, format)
+		},
+		Args: cobra.ExactArgs(1),
+	}
+)
+
+// Execute executes the root command.
```
XinruZhang, line 38: Typo?

tools/tkn-results/cmd/results_list.go
```diff
@@ -0,0 +1,42 @@
...
+				fmt.Printf("ListResults: %v\n", err)
+				return err
+			}
+			return printproto(os.Stdout, resp, format)
+		},
+		Args: cobra.ExactArgs(1),
+	}
+)
+
+// Execute executes the root command.
```
XinruZhang, line 38: Same here

tools/tkn-results/cmd/records_list.go
wlynch: Removed.

tools/tkn-results/cmd/results_list.go
wlynch: Removed.

XinruZhang (approved):
/lgtm

Issue #102, opened 2021-03-31
Look into using dynamic clients for type agnostic watching


wlynch:
We have a lot of duplication in the watcher code, mostly because we don't have a generic type to use for TaskRuns and PipelineRuns. 

[Dynamic clients](https://pkg.go.dev/k8s.io/client-go/dynamic) might be a good compromise here to use [Unstructured](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured) types to support anything that is Object-like.

PR #103, opened 2021-03-31
Add support for deleting resources on completed Runs.

### LINKED ISSUE #59: Watcher: Pipeline etcd cleanup

wlynch:
This implements TEP-0052, adding flag gated support for automatically
cleaning up TaskRun and PipelineRun resources on completion. To be
deleted, the following conditions need to be met:

- The `completed_run_grace_period` flag must be non-zero
- The resource must me marked completed.
- The time elapsed from last Result API update to the current time must
be > `completed_run_grace_period`.
- The resource must not have any `ownerReference` defined (we
assume the resource will be deleted by its owner).

If the resource is not ready for deletion, we reschedule a
reconcile in the future to try again.

For now, `completed_run_grace_period` is disabled by default.

Most of the code is duplicative between TaskRuns and PipelineRuns. Opened https://github.com/tektoncd/results/issues/102 to track refactoring this in another PR.

Fixes #59


pkg/watcher/reconciler/taskrun/taskrun_test.go
```diff
@@ -184,3 +188,100 @@ func TestDisableCRDUpdate(t *testing.T) {
...
+			t.Run("delete", func(t *testing.T) {
+				// Clear out OwnerRefs to make the object eligible for deletion.
+				tr.OwnerReferences = nil
+				if _, err := env.pipeline.TektonV1beta1().TaskRuns(tr.GetNamespace()).Update(tr); err != nil {
+					t.Fatalf("Update: %v", err)
+				}
+				// Force the wall clock forward (with some extra buffer) to
+				// simulate the progression of time past the configured grace
+				// period.
+				fakeClock.Advance(tc.gracePeriod + time.Minute)
```
sbwsg, line 270: I'm not clear why the extra buffer is required - did you experience the test racing before this addition?

sbwsg: FWIW not a major concern. But wondering if a buffer size of a minute might have potential to mask future timing bugs that could otherwise be caught?

pkg/watcher/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -63,26 +65,40 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 
-	if a := pr.GetAnnotations(); result.GetName() == a[annotation.Result] && record.GetName() == a[annotation.Record] {
-		// Result annotations are already present in the PipelineRun.
-		// Nothing else to do.
-		return nil
-	}
+	// If the PipelineRun is complete and not yet marked for deletion, cleanup the run resource from the cluster.
+	if pr.IsDone() && r.cfg.GetCompletedResourceGracePeriod() != 0 {
+		if o := pr.GetOwnerReferences(); len(o) > 0 {
+			log.Infof("resource is owned by another object, defering deletion to parent resource(s): %v", o)
```
sbwsg, line 87: I am wondering if this and the Infof message on line 93 might be better suited to a `Debugf` level instead?  I'm not totally sure if someone administering a large CI/CD cluster will get a ton of benefit from messages like this one at the default log level.

Can definitely see keeping the message on line 97 though, since it indicates an impending modification to the cluster.

sbwsg (commented):
Looks great!  I left a couple comments but otherwise it lgtm.

pkg/watcher/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -63,26 +65,40 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+		// We haven't hit the grace period yet - reenqueue the key for processing later.
+		if s := clock.Since(record.GetUpdatedTime().AsTime()); s < r.cfg.GetCompletedResourceGracePeriod() {
+			log.Infof("pipelinerun is not ready for deletion - grace period: %v, time since completion: %v", r.cfg.GetCompletedResourceGracePeriod(), s)
+			r.enqueue(pr.GetNamespacedName(), r.cfg.GetCompletedResourceGracePeriod())
+			return nil
+		}
+		log.Infof("deleting PipelineRun UID %s", pr.GetUID())
+		if err := r.pipelineclientset.TektonV1beta1().PipelineRuns(pr.GetNamespace()).Delete(pr.Name, &metav1.DeleteOptions{}); err != nil {
+			log.Errorf("PipelineRun.Delete: %v", err)
+			return err
```
ImJasonH, line 100: Just to confirm: do we expect this to re-enqueue the item in the workqueue so we'll retry with backoff?

pkg/watcher/reconciler/pipelinerun/controller.go
```diff
@@ -43,6 +48,7 @@ func NewControllerWithConfig(ctx context.Context, client pb.ResultsClient, cfg *
 	}
 
 	impl := controller.NewImpl(c, logger, "PipelineRunResultsWatcher")
+	c.enqueue = impl.EnqueueKeyAfter
```
ImJasonH, line 51: Any reason to prefer `EnqueueKeyAfter` instead of `EnqueueAfter` which can just take the whole object?

pkg/watcher/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -63,26 +65,40 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-		log.Errorf("PipelineRun.Patch: %v", err)
-		return err
+		// We haven't hit the grace period yet - reenqueue the key for processing later.
+		if s := clock.Since(record.GetUpdatedTime().AsTime()); s < r.cfg.GetCompletedResourceGracePeriod() {
+			log.Infof("pipelinerun is not ready for deletion - grace period: %v, time since completion: %v", r.cfg.GetCompletedResourceGracePeriod(), s)
+			r.enqueue(pr.GetNamespacedName(), r.cfg.GetCompletedResourceGracePeriod())
+			return nil
+		}
+		log.Infof("deleting PipelineRun UID %s", pr.GetUID())
+		if err := r.pipelineclientset.TektonV1beta1().PipelineRuns(pr.GetNamespace()).Delete(pr.Name, &metav1.DeleteOptions{}); err != nil {
```
ImJasonH, line 98: This should handle `k8serrors.IsNotFound` as a no-op, in case we have dueling deletes.

ImJasonH, line 98: wdyt about setting [`DeleteOptions.Preconditions.UID`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#DeleteOptions) to avoid possible mixups if two runs end up having the same name?

It'd be extra confusing to debug this since the previous line logs "deleting UID X" and might end up deleting something else under the wrong circumstances.

pkg/watcher/results/results.go
```diff
@@ -118,7 +120,8 @@ func resultName(o metav1.Object) string {
 	return result.FormatName(o.GetNamespace(), part)
 }
 
-// upsertRecord updates or creates a record for the object
+// upsertRecord updates or creates a record for the object. If there has been
+// no change in the Record data, the existing Record is returned.
```
ImJasonH, line 124: This change seems unrelated to deleting completed runs, was this intentional?

pkg/watcher/reconciler/taskrun/taskrun_test.go
wlynch: Yeah, it's because we calculate the grace period w.r.t. the Record's last update time, and since this package is external to the Result API, we don't have the same control over the API server's internal wall clock that we do in the API Server unit tests to fake things out to be deterministic.

Because we're mixing real clocks and fake clocks, we need to make sure that we're not stuck in the past when we try and compute the grace period during reconcile.
Instead this takes advantage of our control of the watcher's internal wall clock, fast forwarding to a time in the future that we can guarantee is 1) not before the record last update time and 2) past the grace period for grace period values > 0.

I think you're right that this still has a risk of racing if the time taken until this point is more than a minute (unlikely, but possible), so I added an extra bit to make sure the advancement is w.r.t. the current wall clock - it's now basically a sleep that happens instantaneously. 
As an alternative we could fast forward the clock w.r.t. the Record timestamp, but this probably isn't necessary since all the operations here are synchronous, so time.Now() works as an approximation.

tl;dr we're time traveling 🧙‍♂️

pkg/watcher/reconciler/pipelinerun/pipelinerun.go
wlynch: Yes. See https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

[knative/pkg follows similar behavior](https://github.com/knative/pkg/blob/87426b54db6f1200e2cb9086055fdf7fe90307dc/controller/controller.go#L395), requiring you to wrap the error in [NewPermanentError](https://pkg.go.dev/github.com/knative/pkg/controller#NewPermanentError) to stop the key from being re-enqueued.

wlynch:
/hold 

I'm making mistakes not copying changes 1:1 between the PipelineRun and TaskRun files. Let's hold this in favor of https://github.com/tektoncd/results/pull/104 to consolidate the reconciler handling.

pkg/watcher/results/results.go
ImJasonH: Nice, thanks for that explanation! Agree about putting this context into the PR description, and probably in a doc comment in here too.

ImJasonH (approved):

PR #104, opened 2021-04-01
Use dynamic client to consolidate TaskRun and PipelineRun logic.

### LINKED ISSUE #102: Look into using dynamic clients for type agnostic watching

wlynch:
Most of the logic between TaskRun and PipelineRuns was near identical,
but since we were using typed clients there was no easy way to handle
these with a single library. This PR changes the implementation to use a
dynamic client, changing the controllers to use the same code that
target different resource versions.

The goal is to use this to make it easier to add additional types (e.g.
v1alpha1) without massive duplication of logic for each type.

Fixes #102


pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -12,52 +12,60 @@
...
 	log.With(zap.String("key", key))
 
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		log.Errorf("invalid resource key: %s", key)
 		return nil
 	}
-
+	log.Infof("key: %s, namespace: %s, name: %s, gvr: %v", key, namespace, name, r.gvr)
 	// Lookup current TaskRun.
```
ImJasonH, line 56: This comment is out of date now.

pkg/watcher/reconciler/reconciler_test.go
```diff
@@ -57,102 +62,108 @@ func TestController(t *testing.T) {
 
 	pipeline := fakepipelineclient.Get(ctx)
 	t.Run("taskrun", func(t *testing.T) {
-		reconcileTaskRun(ctx, t, pipeline)
-	})
-	t.Run("pipelinerun", func(t *testing.T) {
-		reconcilePipelineRun(ctx, t, pipeline)
-	})
+		//reconcileTaskRun(ctx, t, pipeline)
```
ImJasonH, line 65: I think this was unintentional.

ImJasonH (commented):
Say, this should make it easier to watch `Run`s too 😄

sbwsg:
/lgtm

I'm adding a hold to give you a chance to review the other comments prior to merge. 

/hold

pkg/watcher/reconciler/dynamic/dynamic.go
wlynch: Changed references from TaskRun -> Object.

pkg/watcher/reconciler/reconciler_test.go
wlynch: Removed.

wlynch:
> Say, this should make it easier to watch `Run`s too 😄

You've seen through my plan 😎

PR #107, opened 2021-04-13
Add results-release service account.

wlynch:
This service account includes the Workload Identity annotations
(https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to) to map this to a correspond GCP Service Account.

Workload Identity is already enabled for the dogfooding cluster.

### DIFF
--- a/tekton/release-run.yaml
+++ b/tekton/release-run.yaml
@@ -18,7 +18,7 @@ kind: PipelineRun
 metadata:
   generateName: results-release-api-
 spec:
-  serviceAccountName: release-right-meow
+  serviceAccountName: results-release
   pipelineRef:
     name: results-release
   params:
--- a/tekton/serviceaccount.yaml
+++ b/tekton/serviceaccount.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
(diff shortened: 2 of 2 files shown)


afrittoli:
@wlynch it's up to you - we have a number of configs that we deploy automatically from plumbing to the various cluster... so if you want that you can add it to plumbing. In https://github.com/tektoncd/plumbing/tree/main/docs we try to keep documented what we have on clusters in terms of secrets, so that when we need to re-create a cluster we hopefully know what to do.
So you could simply create the service account and document it, or we could automate it's deployment so that it stays up to date with its git definition.

wlynch:
> @wlynch it's up to you - we have a number of configs that we deploy automatically from plumbing to the various cluster... so if you want that you can add it to plumbing. In https://github.com/tektoncd/plumbing/tree/main/docs we try to keep documented what we have on clusters in terms of secrets, so that when we need to re-create a cluster we hopefully know what to do.
> So you could simply create the service account and document it, or we could automate it's deployment so that it stays up to date with its git definition.

Thanks! Let's keep it here for now for locality with the rest of the Result configs / resources. We can work out any automation as needed later.

XinruZhang (approved):
/lgtm

Issue #99, opened 2021-03-18
Use jsonb instead of blob


jchesterpivotal:
### Feature request

In the schema, the bodies of Tekton's resources are stored as `blob` types. Blobs are not suitable for indexing and do not allow easy querying by analysts or tools accessing the database instead of the gRPC service.

Switching to `jsonb` would allow for more efficient storage, make it possible to index records and enable rich querying via JSON operators.

wlynch:
Thanks for the FR!

We haven't had a chance to look into indexing support, but this looks like a strong contender. I plan to spend some time this week looking into `jsonb` + and other alternatives. Stay tuned!

wlynch:
Overdue follow up: jsonb seems reasonable, but since this will be a database migration we'll also need to provide tooling for migration. I'll add this to our next milestone!

Issue #120, opened 2021-08-04
JSON Serialized Results


wlynch:
Tracking bug for implementing [TEP-0072 - JSON Serialized Results](https://github.com/tektoncd/community/blob/main/teps/0072-results-json-serialized-records.md)

Issue #101, opened 2021-03-30
Params of Pipeline is not found in Records.Data


vvvsrx:
# Expected Behavior
```
"spec": {
    "params": [
        {
            "name": "gitrevision",
            "value": "9de276f1bb9078c967358ec7aa6fb1a8e8a56b29"
        },
        {
            "name": "gitrepositoryurl",
            "value": "git@github.com:tektoncd/results.git"
        },
        {
            "name": "homepage",
            "value": "https://github.com/tektoncd/results.git"
        }
    ],
    "pipelineRef": {
        "name": "pipeline-XXX"
    },
    "serviceAccountName": "robot",
    "timeout": "1h0m0s",
    "workspaces": [
        {
            "name": "sonar-vault",
            "secret": {
                "secretName": "sonar-auth"
            }
        }
    ]
}
```

# Actual Behavior

None

# Steps to Reproduce the Problem

1.
2.
3.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5", GitCommit:"6b1d87acf3c8253c123756b9e61dac642678305f", GitTreeState:"clean", BuildDate:"2021-03-18T01:10:43Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.12", GitCommit:"7cd5e9086de8ae25d6a1514d0c87bac67ca4a481", GitTreeState:"clean", BuildDate:"2020-11-12T09:11:15Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

   Tekton version : v0.21.0

wlynch:
Thanks for the report!

This looks like we're dropping some fields when storing the object. We should look into generating the spec from the upstream source.

wlynch:
Update on this - I've been looking into upstream proto generation for the pipeline types so that we can use that as the source of truth. I've been able to mostly get this mostly working, but the URL field in knative/pkg statuses is causing issues.

I'm trying to avoid forking these libraries if possible, so I opened https://github.com/knative/pkg/issues/2099 for more discussion.

[...]

PR #121, opened 2021-08-04
JSON Serialized Records + Postgres impl.

wlynch:
Sorry for the large PR here 🙏  It was difficult to break up the API changes from the DB changes unless we want to go through multiple iterations of migrations. If I can do anything to make this review easier just let me know!

---

This commit implements JSON serialized Records
https://github.com/tektoncd/community/blob/main/teps/0072-results-json-serialized-records.md.
This changes the format of the stored Record data from an google.protobuf.Any
proto to our own proto type that expects JSON serialized data. This is
intended to:

1. Make it easier for us to store complete Tekton Pipeline data
payloads - Tekton proto type generation was difficult, so we were
relying on manually created types that missed fields. These fields would
be dropped from the CRD -> proto conversion causing us to lose data.
This change lets us store the complete payload as-is without any loss.
2. Make it easier for clients to provide custom types without needing to
define additional proto types.

Along with this change, we are changing the underlying database from
MySQL -> Postgres in order to take advantage of json specific types/features
(https://www.postgresql.org/docs/9.4/datatype-json.html). This commit
does **not** include any mechanism to migrate previous Records / Results to
Postgres. Tooling for this will be a blocker before the 0.1.3 release.

- Records now use a different Any type that expects JSON serialized
record data.
- List syntax for results / records is now slightly different - you no
longer provide leading `result` or `record` prefixes. e.g.

  Before                             | After
  -----------------------------------|------
  result.name == "foo"               | name == "foo"
  record.data.metadata.name == "bar" | data.metadata.name = "bar"

---

Fixes https://github.com/tektoncd/results/issues/99, https://github.com/tektoncd/results/issues/101, https://github.com/tektoncd/results/issues/120


cmd/api/main.go
```diff
@@ -47,10 +47,12 @@ func main() {
...
 	}
-	// Connect to the MySQL database.
-	// DSN derived from https://github.com/go-sql-driver/mysql#dsn-data-source-name
-	dbURI := fmt.Sprintf("%s:%s@%s(%s)/%s?parseTime=true", user, pass, os.Getenv("DB_PROTOCOL"), os.Getenv("DB_ADDR"), os.Getenv("DB_NAME"))
-	db, err := gorm.Open(mysql.Open(dbURI), &gorm.Config{})
+	// Connect to the database.
+	// DSN derived from https://pkg.go.dev/gorm.io/driver/postgres
+
+	dbURI := fmt.Sprintf("host=%s user=%s password=%s dbname=%s port=5432", os.Getenv("DB_ADDR"), user, pass, os.Getenv("DB_NAME"))
+	//dbURI := fmt.Sprintf("%s:%s@%s(%s)/%s?parseTime=true", user, pass, os.Getenv("DB_PROTOCOL"), os.Getenv("DB_ADDR"), os.Getenv("DB_NAME"))
```
sbwsg, line 54: suggest removing or leaving a comment explaining the alternate usage

pkg/api/server/v1alpha2/records_test.go
```diff
@@ -22,20 +22,22 @@ import (
...
 	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/results/pkg/api/server/db/pagination"
 	"github.com/tektoncd/results/pkg/api/server/test"
 	"github.com/tektoncd/results/pkg/api/server/v1alpha2/record"
 	recordutil "github.com/tektoncd/results/pkg/api/server/v1alpha2/record"
 	"github.com/tektoncd/results/pkg/api/server/v1alpha2/result"
 	resultutil "github.com/tektoncd/results/pkg/api/server/v1alpha2/result"
-	"github.com/tektoncd/results/pkg/internal/protoutil"
+	"github.com/tektoncd/results/pkg/internal/jsonutil"
```
sbwsg, line 32: nit: given jsonutil only holds test helpers I am curious why we store it under `pkg/internal/` vs `test/`

pkg/watcher/convert/convert.go
```diff
@@ -19,66 +19,37 @@ limitations under the License.
...
+	// To get around this we append API Version + Kind
+	// (e.g. tekton.dev/v1beta1.TaskRun).
+	v, k := in.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+	return &rpb.Any{
+		Type:  fmt.Sprintf("%s.%s", v, k),
+		Value: b,
+	}, nil
 }
 
+/*
```
sbwsg, line 52: Similarly here, suggest removing or leaving a comment explaining the reason for keeping it.

sbwsg (approved):
Not sure exactly where but suggest documenting the reason for the configmap-from-sql generation and `/docker-entrypoint-initdb.d` mount. I figured it out after looking up the postgres image on docker hub but it confused the heck out of me initially.

Otherwise not too much to nitpick, looks great!

cmd/api/main.go
wlynch: Removed (this was the old mysql format).

pkg/watcher/convert/convert.go
wlynch: Removed.

config/201-sql-deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+    matchLabels:
+      app: postgres
+  template:
+    metadata:
+      labels:
+        app: postgres
+    spec:
+      containers:
+      - name: postgres
+        image: postgres:latest
```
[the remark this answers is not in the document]
wlynch: Good catch. Pinned to 13.3.

PR #139, opened 2021-11-12
tkn-results: Add viper configuration file.

wlynch:
This adds a viper config file that allows for config options to be set
in files so that they do not need to be specified each time in envvars /
flags.

Currently this only supports the existing config values, but we wish to
add additional values on top of this (e.g. auto-fetch service account
tokens for auth).

This change additionally bumps the min Go version to 1.17 so that we can
take advantage of new t.Setenv features - https://golang.org/doc/go1.17#testing


tools/tkn-results/internal/config/config.go
```diff
@@ -0,0 +1,91 @@
...
+	// Config should be evaluated in the following order (last wins):
+	// 1. Environment variables
+	// 2. Config File
+	// 3. Flags
+
+	// Initial config is contains the env variables,
+	// so that the unmarshal can take priority if those values are set.
+	cfg := &Config{
+		SSL: SSLConfig{
+			RootsFilePath:      viper.GetString("TKN_RESULTS_SSL_ROOTS_FILE_PATH"),
```
sbwsg, line 60: given env var names like `TKN_RESULTS_SSL_ROOTS_FILE_PATH` appear multiple times, suggest adding them as named `const` declarations for DRY.

wlynch: Done!

sbwsg:
Awesome, much nicer to have the config logic extracted.

/lgtm

/approve

PR #140, opened 2021-11-15
Update to Tekton 0.29, Kubernetes 0.21

wlynch:
This is a dependency bump to primarily pull in the latest Tekton and
Kubernetes client versions.

This fixes a go.mod issue where we were previously pulling in
k8s.io/client-go@v11, which was causing issues for downstream package
modules.

Other file changes are cleanups for new API differences (context
clients, knative differences, etc) - the behavior should be identical as
before.


dibyom (approved):
/lgtm

PR #144, opened 2021-11-24
Docs: Fix front-matter formatting.

wlynch:
The spacing in the front-matter comments wasn't playing nice with
prettier, leading to formatting issues. This fixes it for all pages,
and reformats the pages with prettier to be sure.


afrittoli (commented):
Thank you!
/lgtm

afrittoli (commented):
I tried this branch locally with the latest PR on website, and there are still a couple of issues.
The config in the website sync is set to pickup all markdown files, however only README.md has front matter defined.  We should either add front matter to them or exclude them from the sync.

I would recommend providing only the linkTitle for the API and Watchers README, as providing both does not render well:

image

wlynch:
Removed `title` frontmatter. All other changes to be done in https://github.com/tektoncd/website/pull/230

afrittoli (commented):
/lgtm

enarha:
/kind documentation

PR #146, opened 2022-01-06
Use kustomize labels instead of commonLabels.

### LINKED ISSUE #145: Unable to upgrade versions due to immutable labels
# Expected Behavior

Can safely upgrade between release versions. 

# Actual Behavior

```
Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```

# Additional Info

This is being cause by kustomize being too aggressive with the [commonLabels field](https://github.com/tektoncd/results/blob/39f9c8d4e60bdb5210c3933dc77d93e59317559d/config/kustomization.yaml#L25) - this is adding the version label to selectors and other fields that are intended to be immutable.

wlynch:
Common labels sets a ton of labels by default [1] that are additive and
can't be disabled [2]. This PR switches to using the labels field
[3] to disable labelling selectors an only label metadata and
template/metadata fields.

This field is fairly new and isn't documented in the kustomize docs
yet, but seems to be the recommended solution moving forward [4] (also
docs appear to be incoming).

[1] https://github.com/kubernetes-sigs/kustomize/blob/f61b075d3bd670b7bcd5d58ce13e88a6f25977f2/api/konfig/builtinpluginconsts/commonlabels.go
[2] https://github.com/kubernetes-sigs/kustomize/issues/817
[3] https://github.com/kubernetes-sigs/kustomize/blob/10026758d314920e8fa3c9c526525d8577d39617/api/types/labels.go
[4] https://github.com/kubernetes-sigs/kustomize/issues/1009

Fixes #145

### DIFF
--- a/config/kustomization.yaml
+++ b/config/kustomization.yaml
@@ -20,7 +20,22 @@ resources:
 - api.yaml
 - default-clusterroles.yaml
 - watcher.yaml
-commonLabels:
-  app.kubernetes.io/part-of: tekton-results
-  app.kubernetes.io/version: devel
+labels:
+  - pairs:
+      app.kubernetes.io/part-of: tekton-results
+  - pairs:
+      app.kubernetes.io/version: devel
+    includeSelectors: false
+    fields:
+      # Include labels in template fields that are filtered out by includeSelectors.
+      # These are additive, derived from the kustomize commonLabel default.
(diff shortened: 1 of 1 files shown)


sbwsg (approved):

enarha:
/kind misc

PR #153, opened 2022-01-31
configure postgres to use the volume mount for data persistence

Fabian-K:
Hey,

By default postgres uses `/var/lib/postgresql/data` for persistence. Because this is not backed by the PV (mounted to `/var/data`), all data is lost when restarting the postgres pod. This is rather unfortunate 😉

This PR changes this by defining PGDATA.

Best regards,
Fabian


wlynch (approved):
Thanks! :D

PR #154, opened 2022-02-03
fix: add 'list' verb to the tekton-results-watcher clusterrole

goldmann:
This makes it possible to list records via the CLI as described in the development doc.


vdemeester (commented):
SGTM 👼🏼

wlynch:
I don't think we want to add this - though I understand why this would be useful!

The developer documentation for debugging abuses the fact that the cluster watcher has *most* of the permissions necessary to query results, but I think it's actually a feature that the watcher can't list all results - i.e. it can only operate on IDs that it knows about via reconcile.

I think what we should do instead is recommend devs to create a new service account and grant the [`tekton-results-readonly` cluster role](https://github.com/tektoncd/results/blob/717eb6eb810cc2e3281b4eb929ea1d9a37afdb10/config/default-clusterroles.yaml#L15-L22), which has the right permissions for full-query access.

goldmann:
@wlynch Sure, in this case I can update the development documentation to reflect this.

goldmann:
I just updated the documentation. Let me know if this is fine.

wlynch (approved):
/lgtm

Thanks! 🎉

enarha:
/kind documentation

PR #156, opened 2022-02-08
fix(cli): fixes handling of flags

### LINKED ISSUE #155: Cannot specify `tkn results records list` parameters (and probably on other commands)
# Expected Behavior

It should be possible to specify parameters (for example filters) when querying via tkn-results.

# Actual Behavior

```
❯ tkn results records list --help
List Records

Usage:
  tkn-results records list [flags] <result parent>

  <result parent>: Result parent name to query. This is typically "<namespace>/results/<result name>", but may vary depending on the API Server. "-" may be used as <result name> to query all Results for a given parent.

Flags:
  -f, --filter string   CEL Filter
  -h, --help            help for list
  -l, --limit int32     number of items to return. Response may be truncated due to server limits.
  -o, --output string   output format. Valid values: tab|textproto|json (default "tab")
  -p, --page string     pagination token to use for next page

Global Flags:
  -a, --addr string        Result API server address
  -t, --authtoken string   authorization bearer token to use for authenticated requests
❯ tkn results records list -f "some-filter" "default/results/-"
unknown shorthand flag: 'f' in -f
Usage of /home/goldmann/.config/tkn/plugins/tkn-results:
  -a, --addr string        Result API server address
  -t, --authtoken string   authorization bearer token to use for authenticated requests
unknown shorthand flag: 'f' in -f
```

In the first run, it correctly printed correct flags, but when trying to set these -- it fails with a message that suggests it is considering only parent flags. In other words - subcomand flags are being ignored.

1. Install latest version of the tkn-results (0.4.0)
2. Try to specify a flag.
3. See it failing as above.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

```
❯ kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.3", GitCommit:"816c97ab8cff8a1c72eccca1026f7820e93e0d25", GitTreeState:"clean", BuildDate:"2022-01-25T21:25:17Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:34:54Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

```
❯ tkn version
Client version: 0.21.0
Pipeline version: v0.32.1
```

goldmann:
* Adds permament flags
* Moves handling of flags to cmd

Fixes #155

DISCLAIMER: I'm *not* a golang developer and I learn as I go. This is my first golang contribution  ever, so, well, keep that in mind. Feel free to point any problems related to my proposal here.


tools/tkn-results/cmd/records_list.go
```diff
@@ -35,9 +29,12 @@ var (
...
+
+	listFlags(cmd.Flags())
+
+	return cmd
+}
 
 func init() {
-	listFlags(recordsListCmd.Flags())
-	recordsCmd.AddCommand(recordsListCmd)
+	recordsCmd.AddCommand(listRecordsCommand())
```
vdemeester, line 39: nit: Ideally we wouldn't even have this `init` and we would add the command on usage, like [here](https://github.com/tektoncd/cli/blob/main/pkg/cmd/root.go#L85) but that can be done in a follow-up 😉

tools/tkn-results/cmd/root.go
```diff
@@ -21,27 +22,56 @@ import (
...
+		return err
 	}
-	return f.Client(ctx)
+
+	return nil
+}
+
+func init() {
+	RootCmd.PersistentFlags().StringVarP(&addr, "addr", "a", "", "Result API server address")
+	RootCmd.PersistentFlags().StringVarP(&authtoken, "authtoken", "t", "", "authorization bearer token to use for authenticated requests")
```
vdemeester, line 74: Same goes for those, could be set on a func that returns the root command and not be part of `init`

vdemeester (commented):
SGTM, except I think we should drive away from `init` function and do something closer to what we do in cli (https://github.com/tektoncd/cli/blob/main/pkg/cmd/root.go for example)

tools/tkn-results/cmd/records_list.go
goldmann: I'll fix this straight away, thanks!

vdemeester (approved):
Looks way cleaner to me 🤗 Thanks @goldmann

tools/tkn-results/cmd/root.go
```diff
@@ -23,77 +17,39 @@ import (
...
-		if _, err := w.Write(b); err != nil {
-			return err
-		}
-	default:
-		return fmt.Errorf("unknown output format %q", format)
-	}
-	return nil
+// Execute executes the root command.
+func Execute() error {
+	return Root().ExecuteContext(context.Background())
```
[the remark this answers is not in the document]
goldmann: Good question, this is how it was structured before as well. I just changed the invocation here. Either case, I've moved it now!

wlynch (approved):
Sorry for not getting to this sooner. Looks great - thanks! 😃 

/lgtm

Issue #147, opened 2022-01-06
Result Summaries


wlynch:
### Feature request

Tracking issue for implementing [TEP-0088 - Result Summaries](https://github.com/tektoncd/community/blob/main/teps/0088-result-summaries.md)

PR #157, opened 2022-02-11
Refactor reconcilers to use Object interfaces.

wlynch:
This is a prereq for
https://github.com/tektoncd/community/blob/main/teps/0088-result-summaries.md.

This change refactors the reconcilers in order to access objects within Runs
in a generic way using interfaces. Previously we were using
unstructured.Unstructured, which doesn't allow us to use an interface to
access common fields across different Run types (in particular - for
result summaries we need access to the Knative Statuses).

This change should not change any behavior.

Part of #147


dibyom (approved):
Looks good - unstructured.Unstructured only allows access to common metadata fields though now we need access to status fields as well - is that why we need this?

wlynch:
> Looks good - unstructured.Unstructured only allows access to common metadata fields though now we need access to status fields as well - is that why we need this?

Yup exactly. The next step is to use the new Object type to extract the status to extract common fields like timestamps, status, etc.

```yaml
type Object interface {
	metav1.Object
	runtime.Object
	StatusConditionGetter
}
```

khrm (commented):
/remove-kind feature
/kind cleanup

PR #160, opened 2022-02-14
e2e: Update client-go calls to use ctx based signatures.

wlynch:
This should have been updated when we bumped the API version awhile
back, but was missed. Good news this is an easy fix and this code hasn't
bitrotted too badly. Bad news is CI missed this and will require
additional investigation.

Also updates README to use correct test command in Quickstart.


khrm (commented):
/kind cleanup
/remove-kind bug

PR #163, opened 2022-03-02
Add Uber/Automaxprocs to autoset GOMAXPROCS to CPU quota

khrm:
Go runtime is unaware of CPU quota which means it will set GOMAXPROCS
to underlying host vm node. This high value means that GO runtime
scheduler assumes that it has more threads and does context switching
when it might work with fewer threads.
This doesn't happen with our other controllers and services because
sharedmain already import this package for them (that's why it's there in
[go.sum](https://github.com/tektoncd/results/blob/9312200317e70cc33f09a7e674c16915c118b835/go.sum#L1161)) .

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -24,6 +24,14 @@ import (
 	"net/http"
 	"os"
 
+	// Go runtime is unaware of CPU quota which means it will set GOMAXPROCS
+	// to underlying host vm node. This high value means that GO runtime
+	// scheduler assumes that it has more threads and does context switching
+	// when it might work with fewer threads.
+	// This doesn't happen# with our other controllers and services because
+	// sharedmain already import this package for them.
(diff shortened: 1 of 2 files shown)


cmd/api/main.go
```diff
@@ -24,6 +24,8 @@ import (
 	"net/http"
 	"os"
 
+	_ "go.uber.org/automaxprocs"
```
wlynch, line 27: Could you also add the PR description as a comment here?

wlynch (changes requested):
LGTM! Just want to document the reasoning behind the need for this here so that future authors why this is here and why it shouldn't be removed.

cmd/api/main.go
khrm: Done

wlynch (approved):
/lgtm

PR #167, opened 2022-03-22
Add results-info ConfigMap and supporting roles

vinamra28:
This commit adds a `results-info` ConfigMap which contains the version
fo Results installed on the cluster. This is primarily meant to be read
by external tools such as `tkn`.

Adding role and rolebindings so that any system authenticated user can
query this configmap to check the version.

Signed-off-by: vinamra28 <jvinamra776@gmail.com>

### DIFF
--- a/config/config-info.yaml
+++ b/config/config-info.yaml
@@ -1,0 +1,63 @@
+# Copyright 2022 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


vdemeester (approved):

config/config-info.yaml
```diff
@@ -0,0 +1,63 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: info
```
dibyom, line 18: shouldn't this be called results-info?

dibyom: The role/role-binding names should probably also have a results prefix right?

vinamra28: https://github.com/tektoncd/results/blob/fc6adeee8f19d239ad7eab107fe5e187add2bed2/config/kustomization.yaml#L41
is already handling this hence I didn't add any prefix. Hope that's fine?

wlynch (approved):
/lgtm

enarha:
/kind misc

PR #169, opened 2022-03-29
Exercising e2e tests to see what's missing or broken

sbwsg:
This commit is initially just probing to see how the e2e tests break and if there's any route to getting them running ASAP without Kind.


afrittoli:
@sbwsg I'm trying to get pipelines tests running with kind - so I wonder what is the motivation for moving off kind instead?

Issue #184, opened 2022-04-15
panic: runtime error: invalid memory address or nil pointer dereference


joejstuart:
# Expected Behavior
tekton-results-api deploys without an issue

# Actual Behavior
The tekton-results-api pod goes into CrashLoopBackOff status with logs

```
2022/04/15 11:23:41 maxprocs: Leaving GOMAXPROCS=6: CPU quota undefined
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x16ecd97]

goroutine 1 [running]:
database/sql.(*ColumnType).DatabaseTypeName(...)
	database/sql/sql.go:3172
gorm.io/gorm/migrator.ColumnType.DatabaseTypeName(...)
	gorm.io/gorm@v1.23.1/migrator/column_type.go:44
gorm.io/driver/postgres.Migrator.AlterColumn.func1.1(0xc0001cfba0?)
	gorm.io/driver/postgres@v1.3.3/migrator.go:251 +0xf7
gorm.io/gorm.(*DB).Connection(0x18aaf20?, 0xc00065ef50)
	gorm.io/gorm@v1.23.1/finisher_api.go:542 +0x102
gorm.io/driver/postgres.Migrator.AlterColumn.func1(0xc0006faa80)
	gorm.io/driver/postgres@v1.3.3/migrator.go:249 +0x26b
gorm.io/gorm/migrator.Migrator.RunWithValue({{0xe0?, 0xc0006f84b0?, {0x1dd4138?, 0xc0003e3f80?}}}, {0x1a51260?, 0xc000188180}, 0xc00065f390)
	gorm.io/gorm@v1.23.1/migrator/migrator.go:52 +0x13c
gorm.io/driver/postgres.Migrator.AlterColumn({{{0x0?, 0xc0006f84b0?, {0x1dd4138?, 0xc0003e3f80?}}}}, {0x1a51260?, 0xc000188180?}, {0x1705cc9?, 0x0?})
	gorm.io/driver/postgres@v1.3.3/migrator.go:237 +0xb1
gorm.io/gorm/migrator.Migrator.MigrateColumn({{0x8b?, 0xc0006f82d0?, {0x1dd4138?, 0xc0003e3f80?}}}, {0x1a51260, 0xc000188180}, 0xc000000b40, {0x1dd8cd0, 0xc0006e31e0})
[...]

PR #185, opened 2022-04-16
Upgrade postgres to v1.3.4

joejstuart:
The postgres driver was throwing error:
panic: runtime error: invalid memory address or nil pointer dereference
See: https://github.com/tektoncd/results/issues/184


wlynch (approved):
/lgtm

Issue #198, opened 2022-06-17
Persist and Serve TaskRun Logs


adambkaplan:
### Feature request

Enhance Results to do the following:

1. Store `TaskRun` step logs.
2. Provide an API where logs for each `TaskRun` step can be served.
3. Use the existing Results RBAC controls to ensure logs are only served to those who have been granted the right permissions.

### Use case

CI/CD users expect to view the full logs of any given step in a build/pipeline process.
This is primarily driven by two use cases:

1. Debugging a failed build/pipeline.
2. Auditing a build/pipeline process.

This is done in Tekton today by serving the underlying container logs from a `TaskRun` pod. These are stored on the host node and can be lost due to `TaskRun` pruning, cluster maintenance, or other mechanisms that delete the underlying pod.
For auditing purposes, build logs may need to be retained for the life a particular version of software is supported.

The most common means of persisting Kubernetes logs today is with log forwarding tools like [fluentd](https://www.fluentd.org/) and analysis engines like ElasticSearch, Amazon CloudWatch, and Grafana Loki.
These stacks are optimized to stream logs across systems for analysis in real time (this is a good thing!).
They are not built to retain and serve individual log files.

This feature request proposes that the Results watcher and apiserver be extended to store logs for `TaskRun` steps.
These logs can then be fetched by the apiserver from an API endpoint.

adambkaplan:
Credit to @CathalOConnorRH who did a lot of research on our end with fluentd and Loki that led to this feature request.

cc @wlynch - this follows up the "Logs with Tekton Results" item discussed during the Tekton Community Summit.

khrm:
> The most common means of persisting Kubernetes logs today is with log forwarding tools like [fluentd](https://www.fluentd.org/) and analysis engines like ElasticSearch, Amazon CloudWatch, and Grafana Loki. These stacks are optimized to stream logs across systems for analysis in real time (this is a good thing!). They are not built to retain and serve individual log files.
 ELK stack is optimized for storing logs longtime. We are using it in Openshift Logging. At present, we can view pipelineruns and taskruns logs in Openshift Logging. There are some issues with this UX.


> This feature request proposes that the Results watcher and apiserver be extended to store logs for `TaskRun` steps. These logs can then be fetched by the apiserver from an API endpoint.

The problem with storing logs in Postgres/MySQL is that they aren't build for this.


I will start working on this problem next week in two phases. We had a discussion on this in slack but unfortunately, it got lost.

[...]

PR #200, opened 2022-06-28
Added REST proxy over existing RPCs

sayan-biswas:
**What?**
Added REST proxy for the gRPC services.

**Why?**
These changes will allow UI developers to access results API over REST
Context https://github.com/tektoncd/results/issues/198

**How?**
gRPC gateway is used to generate the REST handlers and routing is done through google.api.http annotations in the service definition.


sayan-biswas:
Rebased branch and resolved conflicts with PR [#214](https://github.com/tektoncd/results/pull/214)

dibyom:
hi @sayan-biswas - thanks for assigning. Might take me a few days to get to this :)

cmd/api/main.go
```diff
@@ -134,25 +126,45 @@ func main() {
...
+	// Register gRPC server endpoint for gRPC gateway
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	mux := runtime.NewServeMux()
+	opts := []grpc.DialOption{grpc.WithTransportCredentials(creds)}
+	err = v1alpha2pb.RegisterResultsHandlerFromEndpoint(ctx, mux, ":"+configFile.GRPC_PORT, opts)
+	if err != nil {
+		log.Fatal("Error registering gRPC server endpoint: ", err)
+	}
```
adambkaplan, line 164: As a follow up PR - we should ensure that the api server gracefully terminates if the kubelet needs to stop the container (usually with `SIG_TERM`).

Captured in #229.

cmd/api/main.go
```diff
@@ -134,25 +126,45 @@ func main() {
...
+		log.Fatal("Error loading TLS certificate: ", err)
+	}
+
+	// Register gRPC server endpoint for gRPC gateway
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	mux := runtime.NewServeMux()
+	opts := []grpc.DialOption{grpc.WithTransportCredentials(creds)}
+	err = v1alpha2pb.RegisterResultsHandlerFromEndpoint(ctx, mux, ":"+configFile.GRPC_PORT, opts)
```
adambkaplan, line 161: What is the advantage of using `RegisterResultsHandlerFromEndpoint`? Are we worried about race conditions with `RegisterResultsHandler`?

cmd/api/main.go
```diff
@@ -134,25 +126,45 @@ func main() {
...
+	if err != nil {
+		log.Fatal("Error registering gRPC server endpoint: ", err)
+	}
+
+	// Start REST proxy server
+	log.Printf("REST server Listening on: %s", configFile.REST_PORT)
+	log.Fatal(http.ListenAndServeTLS(":"+configFile.REST_PORT, path.Join(configFile.TLS_PATH, "tls.crt"), path.Join(configFile.TLS_PATH, "tls.key"), mux))
```
adambkaplan, line 168: Not a bad idea that we serve TLS, but perhaps we only do this if the TLS cert and keys are present?

Configuring mTLS can be a bit of a pain - falling back to plain HTTP may help dev workflows or folks who want to use edge termination.

adambkaplan (commented):
A few comments and suggestions, but overall looks okay.

I would love to have an e2e test, but so far the test suite only checks the annotations from the watcher. We should have black box tests against the apiserver. Captured this in #230

cmd/api/main.go
sayan-biswas: `RegisterResultsHandlerFromEndpoint` is a wrapper over `RegisterResultsHandler`. It includes the gRPC dial method to get the connection, and also closes the connection when the context is done.
Race condition may arise with `RegisterResultsHandler` for edge cases.

cmd/api/main.go
sayan-biswas: Right!! Will add `syscall.SIGTERM` handling.

adambkaplan: Just to note - don't add it in this pull request. The controller also needs to handle graceful termination.

sayan-biswas: Yes. I will open a different PR for this.

cmd/api/main.go
sayan-biswas: Non TLS support added. If TLS files are not present, both gRPC and REST listner will start without TLS.

cmd/api/README.md
```diff
@@ -2,18 +2,21 @@
...
+|----------------------|----------------------------|----------------------------------------------|
+| DB_USER              | Postgres Database user     | user                                         |
+| DB_PASSWORD          | Postgres Database Password | hunter2                                      |
+| DB_HOST              | Postgres Database host     | /cloudsql/my-project:us-east1:tekton-results |
+| DB_NAME              | Postgres Database name     | tekton_results                               |
+| DB_SSLMODE           | Database SSL mode          | verify-full                                  |
+| GRPC_PORT            | gRPC Server Port           | 50051 (default)                              |
+| REST_PORT            | REST proxy Port            | 8080  (default)                              |
+| PROMETHEUS_PORT      | Prometheus Port            | 9090  (default)                              |
+| TLS_OVERRIDE         | Override TLS SAN name      | results.tekton.dev                           |
```
adambkaplan, line 15: Looking at the godoc, this appears to be hostname override, which really should only be used for testing. We should be more descriptive:

```suggestion
| TLS_HOSTNAME_OVERRIDE         | Override the hostname used to serve TLS. This should not be set (or set to the empty string) in production environments.      | results.tekton.dev                           |
```

cmd/api/main.go
```diff
@@ -45,17 +42,20 @@ import (
...
+	DB_USER         string `mapstructure:"DB_USER"`
+	DB_PASSWORD     string `mapstructure:"DB_PASSWORD"`
+	DB_HOST         string `mapstructure:"DB_HOST"`
+	DB_PORT         string `mapstructure:"DB_PORT"`
+	DB_NAME         string `mapstructure:"DB_NAME"`
+	DB_SSLMODE      string `mapstructure:"DB_SSLMODE"`
+	GRPC_PORT       string `mapstructure:"GRPC_PORT"`
+	REST_PORT       string `mapstructure:"REST_PORT"`
+	PROMETHEUS_PORT string `mapstructure:"PROMETHEUS_PORT"`
+	TLS_OVERRIDE    string `mapstructure:"TLS_OVERRIDE"`
```
adambkaplan, line 57: In keeping with the previous comment:
```suggestion
	TLS_HOSTNAME_OVERRIDE    string `mapstructure:"TLS_HOSTNAME_OVERRIDE"`
```

cmd/api/main.go
```diff
@@ -134,25 +129,52 @@ func main() {
...
+		log.Fatalf("Failed to listen: %v", err)
+	}
+	go func() {
+		log.Printf("gRPC server listening on: %s", configFile.GRPC_PORT)
+		log.Fatal(s.Serve(lis))
+	}()
+
+	// Load REST client TLS cert to connect to the gRPC server
+	if tlsError == nil {
+		creds, err = credentials.NewClientTLSFromFile(path.Join(configFile.TLS_PATH, "tls.crt"), configFile.TLS_OVERRIDE)
```
adambkaplan, line 154: Name update per my previous comment:
```suggestion
		creds, err = credentials.NewClientTLSFromFile(path.Join(configFile.TLS_PATH, "tls.crt"), configFile.TLS_HOSTNAME_OVERRIDE)
```

env/config
```diff
@@ -1,6 +1,11 @@
...
-DB_ADDR=
+DB_HOST=
+DB_PORT=5432
 DB_NAME=
-DB_SSLMODE=
+DB_SSLMODE=disable
+GRPC_PORT=50051
+REST_PORT=8080
+PROMETHEUS_PORT=9090
+TLS_OVERRIDE=
```
adambkaplan, line 10:
```suggestion
TLS_HOSTNAME_OVERRIDE=
```

adambkaplan (commented):
/approve

Generally looks good - a suggestion on the `TLS_OVERRIDE` option. Once that is in, please squash commits so we can merge.

dibyom:
> Unsure why Prow is not adding the approved label, since I am in the OWNERS file...

Hmm...this is odd

dibyom:
@adambkaplan https://github.com/tektoncd/community/pull/828 should fix this

adambkaplan (approved):
/lgtm

PR #203, opened 2022-07-05
Proof of Concept: Persist and Serve TaskRun Logs

adambkaplan:
This PR is a proof of concept for persisting and serving TaskRun logs with Tekton Results (#198). In this concept, the Results apiserver receives logs from the watchers, and stores the logs to a persistent volume. These logs can then be retrieved and served over gRPC.

The watcher utilizes the `tektoncd/cli` (tkn) libraries to fetch and forward the logs, preserving the formatting that `tkn` users are familiar with. gRPC streaming APIs are used to ensure that apiservers are not overloaded. When forwarding logs, a separate results record is created to contain relevant metadata used to fetch and serve the logs. 

The apiserver receives logs from the watcher, ensuring that the location of the file and file size are stored in the results database. The logs can then be retrieved by the log record's name. The concept - as it stands - only supports forwarding and serving logs from the mounted persistent volume. Other file serving mechanisms could be supported, such as retrieving logs from Loki or ElasticSearch.

_Note to reviewers:_ This PR is intended to present an overall picture of how #198 could be implemented. I would not be offended if it can't be merged as is, as it is quite large and consists of several commits that can't be easily squashed/reorganized.


afrittoli:
Thank you @adambkaplan !
Is this something you would like to demo in one of our WG?

vdemeester:
> The watcher utilizes the `tektoncd/cli` (tkn) libraries to fetch and forward the logs, preserving the formatting that `tkn` users are familiar with. gRPC streaming APIs are used to ensure that apiservers are not overloaded. When forwarding logs, a separate results record is created to contain relevant metadata used to fetch and serve the logs.

I think this is gonna be a good enough reason to extract the *fetch and forward* the logs part from the `cli`, most likely in upstream `tektoncd/pipeline`, as I'd rather have`tektoncd/results` depend on `tektoncd/pipeline` (or something even lighter) than on `cli` (that essentially brings all the other dependency).

adambkaplan:
> Thank you @adambkaplan ! Is this something you would like to demo in one of our WG?

Absolutely! I've added it to the Pipelines working group meeting.

> I think this is gonna be a good enough reason to extract the fetch and forward the logs part from the cli, most likely in upstream tektoncd/pipeline, as I'd rather havetektoncd/results depend on tektoncd/pipeline (or something even lighter) than on cli (that essentially brings all the other dependency).

+1 to this. In drafting the concept I discovered a few quirks because the code assumes the cli is the only consumer. I can file an issue on tektoncd/pipeline to track this.

khrm:
I think one problem with this approach is that it ignores fluentd. We generally don't use Kube Client API for logging except in CLI.  Also, we would need to accommodate Google Cloud Logging, Splunk, ELK & OpenShift Logging.

adambkaplan:
> I think one problem with this approach is that it ignores fluentd.

+1 to this - plenty of Kube clusters today use fluentd to forward application and system logs. There will need to be some mechanism for a) the apiserver to know where the logs can be fetched (ex - from Elasticsearch), and b) for the watcher to know that it needs to create a TaskRunLog record, but does _not_ need to upload the log.

The PoC has this bit here where the `TaskRunLog` record is created on the apiserver side [1]. What is notable is that the watcher does not know anything about the backend for the logs. Instead, the apiserver sets the log backend type in the "spec" field. This is then sent back to the watcher when the record is created and returned over gRPC.

I think I can do better here with an API that does not expose the backend type in "spec" - instead it just tells the watcher if logs need to be forwarded. Something like this:

```
kind: TaskRunLog
spec:
  logForwarder: None | Results | External
```
where `None` means there are no logs forwarded, `Results` means logs should be forwarded to Results, and `External` means fluentd/another system forwards the logs. The backend type info can then be relegated to the `status` sub-field.

[1] https://github.com/tektoncd/results/pull/203/files#diff-881653a8a974344ec006fcc20a3ec0c992a194d1b789eb4056ee357cf7ad7831R111-R126

adambkaplan:
> Also, we would need to accommodate Google Cloud Logging, Splunk, ELK & OpenShift Logging.

This is the intent of the `TaskRunLogType` - to allow other backends in the future [1]. Maybe one day this logic can live as a true "plug-in" that can deployed as a sidecar container with the apiserver, but for now I would recommend the logic for these backends live "in tree", so to speak.

[1] https://github.com/tektoncd/results/pull/203/files#diff-83680fa39b0bf1807ef48e2ec46f89b5ef89abf69a628ba8fac9a2e01554e473R28-R32

adambkaplan:
Video link to demo: https://www.youtube.com/watch?v=VVcLeEi9NL4

adambkaplan:
/close

This was a proof of concept. More formal code will come now that the TEP has merged.

PR #214, opened 2022-08-02
Allow configuration of API server through config file

CathalOConnorRH:
Prep work for [TEP-0117](https://github.com/tektoncd/community/pull/759/files) to allow future configuration options through config file. 

This change allows for both config file options and env var options, where the env var will override the config file option if it is set.


cmd/api/main.go
```diff
@@ -48,22 +49,51 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+type ConfigFile struct {
+	DB_USER     string `mapstructure:"DB_USER"`
+	DB_PASSWORD string `mapstructure:"DB_PASSWORD"`
```
adambkaplan, line 54: Do we want to allow DB credentials to be stored in plaintext? Will Viper also allow these to be set with env vars, and the other settings to be populated via the file?

cmd/api/main.go
```diff
@@ -48,22 +49,51 @@ import (
...
+
+	configFile := ConfigFile{}
+	err = viper.Unmarshal(&configFile)
+
+	if err != nil {
+		log.Fatal("cannot load config:", err)
+	}
+
 	flag.Parse()
+	err = viper.Unmarshal(&configFile)
```
adambkaplan, line 82: Duplication from line 75?

CathalOConnorRH: remove the duplicate

cmd/api/main.go
CathalOConnorRH: it's optional to have the password in the config file, if a config item is set as a env var, viper will use that over the config file version [1]. 

[1](https://github.com/spf13/viper/blob/master/viper.go#L1399:L1400)

adambkaplan:
A question for the maintainers (@dibyom, perhaps) - is it worth updating the `api.yaml` manifest to mount a ConfigMap with a config file that contains non-sensitive info (example - db name, address, protocol?)

dibyom:
> A question for the maintainers (@dibyom, perhaps) - is it worth updating the api.yaml manifest to mount a ConfigMap with a config file that contains non-sensitive info (example - db name, address, protocol?)

Sounds good as long as we don't allow passwords over config maps

cmd/api/README.md
```diff
@@ -11,6 +11,7 @@
 | DB_NAME              | MySQL Database name             | tekton_results                               |
 | DB_SSLMODE           | Database SSL mode               | verify-full                                  |
 
+These values can also be set in the config file located in the [env](https://github.com/tektoncd/results/env/config) directory.
```
adambkaplan, line 14: This is creating a broken link, causing the build test to fail. The link probably can't be added until this PR merges.

```

----------------------------------------------
---- Checking links in the markdown files ----
----------------------------------------------
cmd/api/README.md
	ERROR	https://github.com/tektoncd/results/env/config
		Not Found (HTTP error 404)
```

CathalOConnorRH: Remove the url and replaced with folder path. I will update the README with the correct link in another PR after this has been merged.

dibyom (approved):
/lgtm

dibyom (approved):

Issue #231, opened 2022-09-12
Unable to customize the result's parent via the result annotation


alan-ghelardi:
# Expected Behavior

When the result's name is set via the results.tekton.dev/result annotation, the watcher should use the parent defined in the annotation value rather than assuming the object's namespace as the parent. 

# Actual Behavior

The watcher always uses the object's namespace, regardless the result annotation value. When the parent isn't the object's namespace, the API returns the following error: error updating Record: rpc error: code = InvalidArgument desc = requested parent does not match resource name. 

# Steps to Reproduce the Problem

1. Create a PipelineRun in an arbitrary namespace (say dev) and annotate it with the results.tekton.dev/result annotation. The parent's name must differ from the namespace (say foo).
2. Follow the watcher's logs. The following error appears: error updating Record: rpc error: code = InvalidArgument desc = requested parent does not match resource name. 

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.11-eks-18ef993", GitCommit:"b9628d6d3867ffd84c704af0befd31c7451cdc37", GitTreeState:"clean", BuildDate:"2022-07-06T18:06:23Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version: Pipeline version: v0.35.1

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

PR #232, opened 2022-09-12
Read the parent name from the results annotation when it's set

alan-ghelardi:
Closes https://github.com/tektoncd/results/issues/231.

Nowadays the Watcher always uses the object's namespace as the parent name
regardless the value of the results annotation. This pull request alters this
behavior by making the watcher to read the parent name from the results
annotation when it's present.


alan-ghelardi:
@dibyom shall I do something to move forward with this pull request?

vdemeester (approved):

khrm (commented):
/kind bug

PR #240, opened 2022-09-25
update & fix links

Arhell:
resolve https://github.com/tektoncd/results/issues/239


AlanGreene:
Those files are from vendored dependencies, i.e. the result of running `go mod vendor` or similar and should be treated as generated files. I'm open to correction but not sure we should be making manual changes to these as the changes will not be preserved.

Issue #243, opened 2022-09-28
API server is no longer starting  up when installed from the main branch


alan-ghelardi:
# Expected Behavior

Users must be able to install the API server from the code in the main branch.

# Actual Behavior

The API server remains in crash loop.

# Steps to Reproduce the Problem

1. Check out the default branch.
2. Create a Kind cluster.
3. Install the project: `export KO_DOCKER_REPO=kind.local && kustomize build config | ko apply -f -`.
4. Inspect the tekton-results-api pod. It starts and completes immediately. Logs do not provide any clue on what is going  on. 

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  (paste your output here)
  ```

Irrelevant

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

Irrelevant

After debugging the issue, I've found two issues:

1. The API server attempts to load a config file during the boot. If the config file isn't mounted, the main function returns without any information: https://github.com/tektoncd/results/blob/main/cmd/api/main.go#L70-L72.
2. Currently, the default manifests do not bundle the configuration declared at the env/config file.

PR #244, opened 2022-09-28
Fix the API server startup

alan-ghelardi:
## What

This pull request resolves https://github.com/tektoncd/results/issues/243, which
provides a detailed explanation about the issue.

## How

* Call log.Fatalf if the config file that the API server depends upon cannot be
found instead of terminating silently to provide more context to
users.
* Generate a ConfigMap from the env/config file and mount it into the API server
to fix the startup error. This simplifies the installation of the API server.

## Heads-up

I moved the env folder to the config directory. Otherwise, people would need to
disable the load restrictions when calling Kustomize. Apparently, this change
doesn't break anything, but let me know if you people prefer a different
approach. At a first glance, keeping the configuration in a static ConfigMap
would be even simpler, but I am not aware of the motivations for keeping the
config in a separate file.


vdemeester (approved):

dibyom:
Thank you @alan-ghelardi 

/lgtm

khrm (commented):
/kind bug

Issue #248, opened 2022-10-07
Allow listing results and records without a prior knowledge of the parent name


alan-ghelardi:
### Feature request

Currently, the API server only supports listing records without specifying the exact result name: `default/results/-`. This requires a prior knowledge of the existing parents in the system.

In various scenarios, listing Results and Records without specifying the parent name (e.g. `ListResults 'parent: "-"'` and `ListRecords 'parent: "-/results/-"'` might be quite useful (more details below).  

### Use case

- Implementing a radiator to show the current statuses (or the statuses in a given interval, say, last 24  hours) of all PipelineRuns in the system regardless their namespaces.
- Supporting similar queries as those provided by kubectl with the --all-namespaces flag.
- Simplifying exploration: users may get started by  listing all Results in the system and then, narrow down their queries to find the desired data.

PR #249, opened 2022-10-07
Enhance the support for reading Results and Records across multiple collections

alan-ghelardi:
## What

This pull request resolves https://github.com/tektoncd/results/issues/248, which
provides a more detailed explanation about the rationale. Essentially, these
changes allow users to list results across parents and list records across
results without a prior knowledge of the parent name.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -140,10 +140,14 @@ Fields supported in `order_by`:
 | `created_time` |
 | `updated_time` |
 
+## Reading results across parents
+
+Results can be read across parents by specifying `-` as the parent name. This is useful for listing all results stored in the system without a prior knowledge about the available parents.
+
 ## Reading Records across Results
 
-Records can be read across Results by specifying `-` as the Result name part
-(e.g. `default/results/-`). This can be used to read and filter matching Records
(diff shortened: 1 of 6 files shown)


adambkaplan (changes requested):
If we are going to use `-` to mean "all namespaces" for the parent, then we need to modify the RBAC auth check [1]. `-` is not a valid namespace, and the best case is that subject access review will fail with an invalid namespace. Worst case - it always succeeds, which would be a security vulnerability.

Ideally we would have an e2e test covering this. I feel that is too much to ask for in this PR.

[1] https://github.com/tektoncd/results/blob/main/pkg/api/server/v1alpha2/auth/rbac.go#L85-L94

adambkaplan (approved):
/approve

vdemeester (approved):
/lgtm

alan-ghelardi:
@adambkaplan  thank you for accepting the pull request. I started writing an E2E test case for that scenario, but I stumbled upon the mtls. I'm working on that and I send a new pull request once it's done.

khrm (commented):
/kind feature

Issue #142, opened 2021-11-17
Permission denied on Record operations for namespaced auth.


wlynch:
# Expected Behavior

Namespaced scoped SA should be able to auth records.

# Actual Behavior

Permission denied. :(

Things work if the user is granted cluster-wide access - this smells like we're not parsing out the namespace correctly so we're making the wrong request.

# Steps to Reproduce the Problem

1. ```yaml
    apiVersion: rbac.authorization.k8s.io/v1
    kind: RoleBinding
    metadata:
      name: result-ro
    subjects:
      - kind: ServiceAccount
        name: default
    roleRef:
      apiGroup: rbac.authorization.k8s.io
      kind: ClusterRole
      name: tekton-results-readonly
    ```
2. ```sh
    $ grpc_cli call --channel_creds_type=ssl --ssl_target=tekton-results-api-service.tekton-pipelines.svc.cluster.local --call_creds=access_token=$(kubectl get secrets -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='result-reader')].data.token}"|base64 --decode) localhost:50051 tekton.results.v1alpha2.Results.ListRecords 'parent: "default/results/4ca70fb5-9cfd-4c0b-bdad-c683c02c7b59"'
    Rpc failed with status code 16, error message: permission denied
    ```

# Additional Info

```yaml
$ k get clusterrole tekton-results-readonly -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: tekton-results-readonly
rules:
- apiGroups:
  - results.tekton.dev
  resources:
  - results
  - records
  verbs:
  - get
  - list
```

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:18:45Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
  ```

[...]

PR #253, opened 2022-10-13
Improve coverage of E2E tests

alan-ghelardi:
Resolves https://github.com/tektoncd/results/issues/142.

## What

- Write E2E test cases for the changes introduced on
https://github.com/tektoncd/results/pull/249. In addition, basic listing
operations were covered.
- Fix a bug in the authorization flow for the ListRecords operation. The parent
passed to the `Check` method was incorrect and if one created a Role to restrict
the access to a single namespace, the API server would always return an
unauthenticated error. The fix is covered in a newlycreated E2E test case.
- Pin Kubernetes v1.22 in the Kind config. By doing so, users may run newer
versions of Kind and the tests will run on a cluster that uses a known
Kubernetes version. Note that Knative no longer supports the Kubernetes version
used by default in the Kind v0.10 (which is mentioned in the E2E documentation).

## Additional information

There are two already existing E2E test cases failing consistently. I'm going to
open a new pull request to fix them to make changes more granular.

### DIFF
--- a/test/e2e/01-install.sh
+++ b/test/e2e/01-install.sh
@@ -30,40 +30,49 @@ kubectl create secret generic tekton-results-postgres --namespace="tekton-pipeli
 
 echo "Generating TLS key pair..."
 set +e
-  openssl req -x509 \
-     -newkey rsa:4096 \
-     -keyout "/tmp/tekton-results-key.pem" \
-     -out "/tmp/tekton-results-cert.pem" \
-     -days 365 \
-     -nodes \
-     -subj "/CN=tekton-results-api-service.tekton-pipelines.svc.cluster.local" \
-     -addext "subjectAltName = DNS:tekton-results-api-service.tekton-pipelines.svc.cluster.local"
+openssl req -x509 \
+        -newkey rsa:4096 \
+        -keyout "/tmp/tekton-results-key.pem" \
+        -out "/tmp/tekton-results-cert.pem" \
+        -days 365 \
+        -nodes \
+        -subj "/CN=tekton-results-api-service.tekton-pipelines.svc.cluster.local" \
+        -addext "subjectAltName = DNS:tekton-results-api-service.tekton-pipelines.svc.cluster.local"
 
-  if [ $? -ne 0 ] ; then
+if [ $? -ne 0 ] ; then
     # LibreSSL didn't support the -addext flag until version 3.1.0 but
     # version 2.8.3 ships with MacOS Big Sur. So let's try a different way...
     echo "Falling back to legacy libressl cert generation"
     openssl req -x509 \
-      -verbose \
-      -config <(cat /etc/ssl/openssl.cnf <(printf "[SAN]\nsubjectAltName = DNS:tekton-results-api-service.tekton-pipelines.svc.cluster.local")) \
-      -extensions SAN \
-      -newkey rsa:4096 \
-      -keyout "/tmp/tekton-results-key.pem" \
-      -out "/tmp/tekton-results-cert.pem" \
-      -days 365 \
-      -nodes \
-      -subj "/CN=tekton-results-api-service.tekton-pipelines.svc.cluster.local"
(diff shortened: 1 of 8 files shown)


test/e2e/kustomize/rbac.yaml
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: all-namespaces-read-access
```
adambkaplan, line 18: Specify the namespace for this service account as well, since we specify namespace in the rolebinding.
```suggestion
  name: all-namespaces-read-access
  namespace: default
```

test/e2e/01-install.sh
```diff
@@ -30,41 +30,50 @@ kubectl create secret generic tekton-results-postgres --namespace="tekton-pipeli
...
-  fi
+fi
 set -e
 kubectl create secret tls -n tekton-pipelines tekton-results-tls --cert="/tmp/tekton-results-cert.pem" --key="/tmp/tekton-results-key.pem" || true
 
 echo "Installing Tekton Results..."
 kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply -f -
 
+echo "Fetching access tokens..."
+tokens_dir=/tmp/tekton-results/tokens
```
adambkaplan, line 69: Ideally this can be a configuration option for the script. This can be addressed in a follow-up PR.

test/e2e/grpc_client.go
```diff
@@ -0,0 +1,90 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+const (
+	certFile                  = "/tmp/tekton-results-cert.pem"
+	apiServerName             = "tekton-results-api-service.tekton-pipelines.svc.cluster.local"
+	apiServerAddress          = "localhost:50051"
+	allNamespacesReadAccess   = "/tmp/tekton-results/tokens/all-namespaces-read-access"
+	singleNamespaceReadAccess = "/tmp/tekton-results/tokens/single-namespace-read-access"
```
adambkaplan, line 39: Likewise, when this is made configurable, we should move these constants near a test main.

adambkaplan (commented):
Thank you so much for this PR! This will definitely help get the ball rolling on having sustainable e2e tests. And ++1 for catching a bug in the process!

I have a few suggestions about the rbac.yaml that is deployed, and a question about the kind configuration yaml. The other suggestions regarding configuring the cert pair locations are optional - IMO those should not block merge and can be captured as a follow-up item.

test/e2e/kind-cluster.yaml
```diff
@@ -16,7 +16,12 @@
...
 nodes:
-- role: control-plane
-- role: worker
-- role: worker
-- role: worker
\ No newline at end of file
+- &node
+  role: control-plane
+  image: kindest/node:v1.22.13@sha256:4904eda4d6e64b402169797805b8ec01f50133960ad6c19af45173a27eadf959
+- <<: *node
```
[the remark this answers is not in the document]
alan-ghelardi: I'm gonna add a comment explaining that. Since I pinned the Kubernetes version by adding the appropriate kindest image and the image should be set in all nodes, to avoid a future mistake (someone bumps the Kubernetes version in the control plane and forgets to do the same in the data plane) I used the merge key strategy.

alan-ghelardi:
@adambkaplan  thank you for the review. I've accepted your suggestions and will make the tokens path configurable in a future pr.

adambkaplan (approved):
/approve

khrm (commented):
/kind cleanup

Issue #254, opened 2022-10-18
Object deletion isn't working properly and E2E tests are broken


alan-ghelardi:
# Expected Behavior

If the completed_run_grace_period watcher flag is greater than 0, objects must be deleted upon completion once the configured grace period elapses. 

# Actual Behavior

This feature seems to be unstable and the existing end-to-end test case fails most of the time.

# Steps to Reproduce the Problem

1. Run end-to-end tests (`./test/e2e/e2e.sh`).
2. The test case that verifies whether objects are cleaned up after completion fails most of the  time. 

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

Irrelevant

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

Irrelevant

PR #255, opened 2022-10-18
Fix resource deletion and refactor watcher

alan-ghelardi:
## What

This pull request resolves https://github.com/tektoncd/results/issues/254. The
following fixes and improvements were made:

- Fix object deletion upon completion: apparently, the main cause of the
instability was the use of the record to determine the time at which the Run
object had been updated. I changed the watcher to use the
`Status.CompletionTime` since this approach seems to be more
reliable. Currently, if the Tekton
pipelines controller resyncs the object and transitions the status, the object
will change, the record will be updated and this will cause discrepancies in the
verification of the grace period expiration.
- The aforementioned change fixes end-to-end tests. Now, the test suite is
passing consistently in my machine (this
[pr](https://github.com/tektoncd/results/pull/253) should be merged as well
since it fixes other issues in the current e2e setup).
- Improve logs emitted by the watcher: favor structured logging since it
simplifies a lot the debugging with log indexer tools such as Splunk and others.
- Refactor the controller's code to make it simpler and hopefully, easier to
reason about. I tried to make things more uniform and meaningful for new
collaborators. One of the most relevant changes was removing the enqueue
function that was being passed around and replacing the approach with the
knative's `controller.NewRequeueAfter` function which returns an error that
tells knative to requeue the key after a given duration.

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -16,18 +16,22 @@ package dynamic
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/jonboulle/clockwork"
+	pipelinev1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/results/pkg/watcher/convert"
 	"github.com/tektoncd/results/pkg/watcher/reconciler"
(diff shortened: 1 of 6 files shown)


vdemeester (approved):

khrm (commented):
/approve

khrm (commented):
/kind bug

PR #264, opened 2022-11-11
add s390x arch support

Basavaraju-G:
This will add support for IBM s390x architecture support


adambkaplan (approved):
/approve

@dibyom it probably makes sense that we include ARM releases as well? This could be a follow-up PR.

dibyom:
> @dibyom it probably makes sense that we include ARM releases as well? This could be a follow-up PR.

Yeah!

khrm (commented):
/kind misc

PR #265, opened 2022-11-18
Added compatibility with priviliged pod security level

realSchoki:
Since tekton installation now enforces the pod security level to be restricted I altered the deployment templates a bit and added the needed security settings.

I also swapped out the postgres image to the one offered by bitnami since they make non-root images. A non-root image is necessary for the restricted pod policy. I did not touched the  the version the keep the change minimal.


realSchoki:
I added the missing email mentioned by the EasyCLA.

PR #268, opened 2022-11-27
Hide sensitive information from pod logs.

AndrienkoAleksandr:
Hide sensitive information from pod logs.

Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/tektoncd/results/pkg/api/server/v1alpha2/auth"
 	_ "go.uber.org/automaxprocs"
+	"go.uber.org/zap"
 	"google.golang.org/grpc/credentials/insecure"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -43,6 +44,7 @@ import (
 	"google.golang.org/grpc/reflection"
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
 )
 
 type ConfigFile struct {
@@ -55,12 +57,12 @@ type ConfigFile struct {
 	GRPC_PORT             string `mapstructure:"GRPC_PORT"`
 	REST_PORT             string `mapstructure:"REST_PORT"`
 	PROMETHEUS_PORT       string `mapstructure:"PROMETHEUS_PORT"`
+	LOG_LEVEL             string `mapstructure:"LOG_LEVEL"`
 	TLS_HOSTNAME_OVERRIDE string `mapstructure:"TLS_HOSTNAME_OVERRIDE"`
 	TLS_PATH              string `mapstructure:"TLS_PATH"`
 }
 
 func main() {
-
 	viper.AddConfigPath("./env")
 	viper.SetConfigName("config")
 	viper.SetConfigType("env")
@@ -79,6 +81,9 @@ func main() {
 		log.Fatal("Cannot load config:", err)
 	}
 
+	log, logConf := getLogger(configFile)
+	defer log.Sync()
+
 	if configFile.DB_USER == "" || configFile.DB_PASSWORD == "" {
 		log.Fatal("Must provide both DB_USER and DB_PASSWORD")
 	}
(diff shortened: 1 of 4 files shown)


khrm (commented):
/ok-to-test

cmd/api/main.go
```diff
@@ -87,10 +88,9 @@ func main() {
 
 	dbURI := fmt.Sprintf("host=%s user=%s password=%s dbname=%s port=%s sslmode=%s", configFile.DB_HOST, configFile.DB_USER, configFile.DB_PASSWORD, configFile.DB_NAME, configFile.DB_PORT, configFile.DB_SSLMODE)
 
-	db, err := gorm.Open(postgres.Open(dbURI), &gorm.Config{})
-	if err != nil {
-		log.Fatalf("Failed to open the results.db: %v", err)
-	}
+	db, err := gorm.Open(postgres.Open(dbURI), &gorm.Config{
+		Logger: logger.Default.LogMode(logger.Silent),
+	})
```
adambkaplan, line 93: Does this mean that the GORM logs are permanently suppressed? If yes, can we use the zap log level to turn these back on? For example - log if "debug" is set.

Also Need to ensure that we catch the error opening the database connection and fail:

```suggestion
	db, err := gorm.Open(postgres.Open(dbURI), &gorm.Config{
		Logger: logger.Default.LogMode(logger.Silent),
	})
	if err != nil {
		log.Fatalf("Failed to open the results.db: %v", err)
	}
```

test/e2e/01-install.sh
```diff
@@ -26,7 +26,7 @@ kubectl apply --filename ${TEKTON_PIPELINE_CONFIG}
 
 echo "Generating DB secret..."
 # Don't fail if the secret isn't created - this can happen if the secret already exists.
-kubectl create secret generic tekton-results-postgres --namespace="tekton-pipelines" --from-literal=POSTGRES_USER=postgres --from-literal=POSTGRES_PASSWORD=$(openssl rand -base64 20) || true
+kubectl create secret generic tekton-results-postgres --namespace="tekton-pipelines" --from-literal=POSTGRESQL_USER=postgres --from-literal=POSTGRESQL_PASSWORD=$(openssl rand -base64 20) || true
```
adambkaplan, line 29: Will this break our e2e tests?

adambkaplan (changes requested):

test/e2e/01-install.sh
adambkaplan: looks like this will undo #270 - please revert this change.

cmd/api/main.go
adambkaplan: @AndrienkoAleksandr following up on our slack conversation - I think for now we can suppress these log messages. I will create a follow up issue to make these logs configurable.

AndrienkoAleksandr: Hello, default log library doesn't have log level feature, so if you don't mind I updated my pr to use zap logger and I made log level configurable using config file...

adambkaplan: That's fine, though I suspect we will need to propagate this logger to other spots in the code.

cmd/api/main.go
```diff
@@ -87,7 +92,11 @@ func main() {
 
 	dbURI := fmt.Sprintf("host=%s user=%s password=%s dbname=%s port=%s sslmode=%s", configFile.DB_HOST, configFile.DB_USER, configFile.DB_PASSWORD, configFile.DB_NAME, configFile.DB_PORT, configFile.DB_SSLMODE)
 
-	db, err := gorm.Open(postgres.Open(dbURI), &gorm.Config{})
+	gormConf := &gorm.Config{}
+	if logConf.Level.Level() == zap.DebugLevel {
+		gormConf.Logger = logger.Default.LogMode(logger.Silent)
+	}
```
adambkaplan, line 98: I think this does the opposite of what we want. Logging higher than "debug" should silence the gorm logger.

adambkaplan:
> the code looks good - just a quick question - what is the sensitive information that we are hiding here? Is that the DB/GORM logs?

Mainly the DB/GORM logs. A few other log messages are also being downgraded from "Info" to "Debug" level.

In addition, #275 will add additional observability in the logs.

adambkaplan:
@AndrienkoAleksandr looks good. Please squash your commits and I can lgtm.

AndrienkoAleksandr:
> @AndrienkoAleksandr looks good. Please squash your commits and I can lgtm.

Sorry, for delay, done.

khrm (commented):
/kind feature

PR #270, opened 2022-11-28
Fix E2E tests deployment

enarha:
Bump Kind cluster version. Fix variables names mismatch.

### DIFF
--- a/config/api.yaml
+++ b/config/api.yaml
@@ -41,12 +41,12 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: tekton-results-postgres
-                  key: POSTGRESQL_USER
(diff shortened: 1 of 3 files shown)


khrm (commented):
/ok-to-test

adambkaplan:
@dibyom @vdemeester @afrittoli I'm trying to determine if the e2e tests are actually running on CI. It looks like they aren't?

adambkaplan:
Thanks, @dibyom. I think we should merge this PR (trusting that @enarha has been able to test locally), then try to update the plumbing repo to run the e2e suite properly. Anything that is broken we'll then fix with a follow-up PR.

/approve

PR #274, opened 2022-12-06
Fix api config location and e2e tests

enarha:
The api config location was inconsistent between the code and deployment manifest. The new location was chosen based on
https://github.com/GoogleContainerTools/distroless/issues/718. Kubernetes v1.25 uses new types of tokens for the service accounts which are not automatically created.
Bump test deployment timeout causing sporadic failures.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -62,6 +62,7 @@ type ConfigFile struct {
 func main() {
 
 	viper.AddConfigPath("./env")
+	viper.AddConfigPath("/env")
 	viper.SetConfigName("config")
 	viper.SetConfigType("env")
 
--- a/config/api.yaml
+++ b/config/api.yaml
@@ -53,7 +53,7 @@ spec:
               value: tekton-results
           volumeMounts:
             - name: config
-              mountPath: /home/nonroot/env
+              mountPath: /env
               readOnly: true
             - name: tls
               mountPath: "/etc/tls"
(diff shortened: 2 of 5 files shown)


adambkaplan (approved):
/approve

All looks good to me.

enarha:
Can I have a review please? I'm adding controllers' e2e tests and I would like to get those fixes in first. Thanks.

enarha:
Added 2 small enhancement.

adambkaplan:
Last call for reviews - @dibyom @alan-ghelardi @khrm able to take a look?

khrm (commented):
Isn't container's workdir /home/nonroot? That's why earlier location was working for api server. New location is also fine.

enarha:
@khrm I think when building with ko (which is what we do for the e2e tests), the binary goes to /ko-app/binary.

cmd/api/main.go
```diff
@@ -62,6 +62,7 @@ type ConfigFile struct {
 func main() {
 
 	viper.AddConfigPath("./env")
```
khrm, line 64: Isn't this line redundant now?

khrm (commented):
/lgtm

cmd/api/main.go
enarha: I couldn't verify an actual release (it's failing on issues fixed in code, but not yet released) which could differ from the deployment used for the e2e tests, so I fixed the e2e test deployment with the least amount of risk. Once there is a deployable release I can verify and remove that line if redundant.

adambkaplan:
/lgtm

Adding per @khrm's comment.

khrm (commented):
/kind cleanup

PR #277, opened 2022-12-08
logs storage service and RPC definitions

sayan-biswas:
This PR separates the service definition of logs from the results in the Proto file. Also the logs service registration is made optional at the API server and the watcher can negotiate the availability of logs service through a reflection client and enable/disable sending logs.


cmd/watcher/main.go
```diff
@@ -58,6 +59,7 @@ var (
 	authToken               = flag.String("token", "", "Authentication token to use in requests. If not specified, on-cluster configuration is assumed.")
 	completedRunGracePeriod = flag.Duration("completed_run_grace_period", 0, "Grace period duration before Runs should be deleted. If 0, Runs will not be deleted. If < 0, Runs will be deleted immediately.")
 	threadiness             = flag.Int("threadiness", controller.DefaultThreadsPerController, "Number of threads (Go routines) allocated to each controller")
+	logsAPI                 = flag.Bool("logs_api", true, "Disable sending logs. If not set, the logs will be sent only if server support API for it")
```
adambkaplan, line 62: We are likely going to need to revisit this logic in the future. I like the fact that watcher will only forward logs if the API server publishes the service for it. There will be future cases (ex - logs forwarded via fluentd/fluentbit) where the watcher won't directly forward data, but will need to publish information as to where the logs landed.

OK to keep this here for now.

pkg/watcher/reconciler/pipelinerun/controller.go
```diff
@@ -28,17 +28,18 @@ import (
...
-func NewControllerWithConfig(ctx context.Context, client pb.ResultsClient, cfg *reconciler.Config) *controller.Impl {
+func NewControllerWithConfig(ctx context.Context, resultsClient pb.ResultsClient, cfg *reconciler.Config) *controller.Impl {
 	informer := pipelineruninformer.Get(ctx)
 	lister := informer.Lister()
 
 	c := &Reconciler{
 		LeaderAwareFuncs: leaderelection.NewLeaderAwareFuncs(lister.List),
-		client:           client,
+		resultsClient:    resultsClient,
+		logsClient:       logs.Get(ctx),
```
adambkaplan, line 42: Non-blocking observation - some things we are wiring through the `Context` object, others we are not. Unsure if/how we can establish some consistency here.

cmd/api/main.go
```diff
@@ -121,6 +122,9 @@ func main() {
 		grpc.UnaryInterceptor(prometheus.UnaryServerInterceptor),
 	)
 	v1alpha2pb.RegisterResultsServer(s, v1a2)
+	if configFile.LOGS_API {
+		v1alpha2pb.RegisterLogsServer(s, v1a2)
+	}
```
adambkaplan, line 127: This makes sense to me - if we want logs to be an opt-in feature, disable the API completely if it is not available.

adambkaplan (approved):
/approve

My only main comment is that in the future, we will need to create mechanisms for the watcher to store metadata of logs forwarded by other systems/services.

pkg/watcher/reconciler/pipelinerun/controller.go
sayan-biswas: All the other clients and informers used in Knative are injected in the controller context (as in other tekton controllers). I guess injected client is the consistent way when working with Knative.

cmd/api/README.md
```diff
@@ -2,19 +2,21 @@
...
+| TLS_HOSTNAME_OVERRIDE | Override the hostname used to serve TLS. This should not be set (or set to the empty string) in production environments. | results.tekton.dev                           |
+| TLS_PATH              | Path to TLS files                                                                                                        | /etc/tls                                     |
 | LOG_LEVEL            | Log level for api server   | INFO                                         |
-| TLS_HOSTNAME_OVERRIDE| Override the hostname used to serve TLS. This should not be set (or set to the empty string) in production environments.     | results.tekton.dev                           |
-| TLS_PATH             | Path to TLS files          | /etc/tls                                     |
+| LOGS_API              | Enable logs storage service                                                                                              | false (default)                              |
+| Environment Variable | Description                | Example                                      |
```
khrm, line 19: Remove this line in the doc.

pkg/watcher/reconciler/pipelinerun/controller.go
khrm: Yes. We can try moving to knative/pkg but it's okay for now. We did this with the Trigger's EventListener sink HTTP server. The first implementation wasn't Knative. I will look into doing this with GRPC.

khrm: Yes. We can try moving to knative/pkg but it's okay for now. We did this with the Trigger's EventListener sink HTTP server. The first implementation wasn't Knative. I will look into doing this with GRPC.

adambkaplan:
/assign @dibyom 

/cc @alan-ghelardi 

Discussed with @sayan-biswas  - I closed #258, and instead of one big PR we think it's better to land it in pieces. This adds logs as an optional feature on the apiserver. If logs are disabled, the APIs are not served.

alan-ghelardi:
Awesome, although I cannot approve yet.

dibyom (commented):
/lgtm

khrm (commented):
/kind feature

Issue #289, opened 2022-12-20
sql-initdb-config ConfigMap is out of date


alan-ghelardi:
The sql-initdb-config ConfigMap contains an outdated version of the schema/results.sql file. For example, the `Summary` related  fields aren't present in the schema what causes the database to be initialized with an  outdated schema when users install the project using the  current config files.

alan-ghelardi:
Fixed through a different approach at https://github.com/tektoncd/results/pull/334
/close

PR #290, opened 2022-12-21
Upgrade sql-initdb-config ConfigMap

alan-ghelardi:
This pull request resolves https://github.com/tektoncd/results/issues/289 by
upgrading the sql-initdb-config ConfigMap to install the latest schema.

## Note

I deleted the generate.go in favor of using Kustomize's ConfigMap
generators. The motivation is avoiding duplication since in the current approach
both files, result.sql and 100-sql-initdb-config.yaml, should be generated and
committed to ensure that the correct schema will be installed on the database
initialization. The new approach seems to be simpler: now only the results.sql
file need to be updated once the database schema changes and the
sql-initdb-config ConfigMap will be generated in runtime once the kustomization
manifest is processed.


dibyom:
(minor) Is it possible to keep the results.sql in the schema folder while still using kustomize configmap generator?

alan-ghelardi:
@dibyom  it is, but Kustomize, by default, forbids the kustomization file from referencing files from outside their root folder. To disable this restriction, users need to pass the `--load-restrictor=LoadRestrictionsNone` while building the kustomization file. Usually, this isn't recommended, that's why I moved the results.sql to the config folder.
Do you think that there're benefits in keeping the file under the schema directory? If so, maybe we could create a symblink from schema/results.sql to config/results.sql?

alan-ghelardi:
@dibyom  any thoughts on that?

adambkaplan:
I am personally OK with keeping the sql in the `config` Kustomization directory. Do we have mechanisms to sync the .sql contents with the GORM model?

dibyom:
hi @alan-ghelardi if we could symlink the file to config, that'd be ideal. But if not, I'm ok moving it to config (do we need to update any docs anywhere?)

alan-ghelardi:
@adambkaplan  and @dibyom I was wondering if we could simply remove this schema and let Gorm run the auto-migration by default, giving the option of disabling it if people prefer. I described this in more details at https://github.com/tektoncd/results/issues/311.
I googled a bit, and didn't find a way of extracting the DDL script from Gorm models. So, I'd suggest removing the file and adjusting the models to reflect the most recent schema. Do you people see any strong argument for keeping this file with the burden of keeping it up to date with the models?

dibyom:
+1 to making the migration optional.

I think we started with the sql file first and then migrated to Gorm. I'm not sure if there is any historical reason for keeping both. I think in the beginning we used SQLite as the DB and later shifted to Postgres - does GORM support jsonb?

adambkaplan:
likewise +1 on ability to disable auto-migration as well. May be nice to have a sub-command (or separate binary) that lets the migrator be run as a `Job` externally from the apiserver.

alan-ghelardi:
I started looking into that more deeply and after making a few adjustments in the models, I managed to get tests working without installing the schema directly and letting Gorm to run the auto-migration. Actually, there are two tests failing (those that verify if the same objects cannot be inserted twice). Maybe some uniqueness constraint is missing. I'm gonna fix it and will open a pr.

dibyom:
I think we can close this out now? in favor of #334 ?

Issue #291, opened 2022-12-22
Results isn't deleting non-succeeded objects


alan-ghelardi:
# Expected Behavior

When configured to do so, Results must delete all completed objects regardless their status.

# Actual Behavior

Only succeeded objects are deleted. Failing objects remain in the cluster indefinitely.

# Steps to Reproduce the Problem

1. Configure Results to delete objects upon completion.
2. Create a PipelineRun that will fail or cancel a running PipelineRun.
3. Note that the corresponding result and records are created properly, but the objects remain in the cluster.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

Irrelevant

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

Irrelevant

The root cause of this issue is a mistaken in the implementation of the isDone function: https://github.com/tektoncd/results/blob/main/pkg/watcher/reconciler/dynamic/dynamic.go#L170-172. It checks if the condition's status is True instead of verifying if it isn't Unknown. I'll submit a fix soon.

PR #294, opened 2022-12-23
Fix isDone function, improve visibility as well as traceability and skip deleted keys

alan-ghelardi:
This pull request resolves https://github.com/tektoncd/results/issues/291.

## What

The function isDone actually is checking if the resource has been succeeded. As
of this fix it'll work as expected, by verifying if the object is completed
regardless its status.

## Additional improvements

- Log that the object is being reconciled after checking if the current instance
is the leader. When multiple replicas of the Watcher are running, this avoids
causing the false impression that the same key is being reconciled by various
instances.
- Skip silently keys that are no longer available. This avoid cluttering logs
with irrelevant errors when an object is deleted by a Watcher instance, but
still exists in the work queue of another one.
- Tweak logs, wrap certain errors and add time tracing to certain parts of the
Watcher's code. These changes aim to improve visibility and traceability for
engineers who are debugging the Watcher.

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -70,18 +70,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, o results.Object) error {
 		logger.Debugf("Post SetGroupVersionKind: %s", o.GetObjectKind().GroupVersionKind().String())
 	}
 
-	// Update record.
+	// Upsert record.
+	startTime := time.Now()
 	result, record, err := r.resultsClient.Put(ctx, o)
+	timeTakenField := zap.Int64("results.tekton.dev/time-taken-ms", time.Since(startTime).Milliseconds())
+
 	if err != nil {
-		logger.Debugw("Error updating Record", zap.Error(err))
-		return err
+		logger.Debugw("Error upserting record", zap.Error(err), timeTakenField)
+		return fmt.Errorf("error upserting record: %w", err)
 	}
 
 	logger = logger.With(zap.String("results.tekton.dev/result", result.Name),
 		zap.String("results.tekton.dev/record", record.Name))
+	logger.Debugw("Record has been successfully upserted into API server", timeTakenField)
 
 	if err := r.addResultsAnnotations(logging.WithLogger(ctx, logger), o, result, record); err != nil {
-		return err
+		return fmt.Errorf("error adding Result annotations to the object: %w", err)
 	}
 
 	return r.deleteUponCompletion(logging.WithLogger(ctx, logger), o)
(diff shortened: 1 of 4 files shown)


pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -156,19 +160,21 @@ func (r *Reconciler) deleteUponCompletion(ctx context.Context, o results.Object)
...
+		return fmt.Errorf("error deleting object: %w", err)
 	}
 
+	logger.Debugw("Object has been successfully deleted", zap.Int64("results.tekton.dev/time-taken-seconds", int64(time.Since(*completionTime).Seconds())))
 	return nil
 }
 
 func isDone(o results.Object) bool {
-	return o.GetStatusCondition().GetCondition(apis.ConditionSucceeded).IsTrue()
+	return !o.GetStatusCondition().GetCondition(apis.ConditionSucceeded).IsUnknown()
```
dibyom, line 177: Yikes...could we add a test for this so this sort of thing does not go undetected?

alan-ghelardi:
>Yikes...could we add a test for this so this sort of thing does not go undetected?
Done @dibyom .

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -70,18 +70,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, o results.Object) error {
 		logger.Debugf("Post SetGroupVersionKind: %s", o.GetObjectKind().GroupVersionKind().String())
 	}
 
-	// Update record.
+	// Upsert record.
+	startTime := time.Now()
 	result, record, err := r.resultsClient.Put(ctx, o)
+	timeTakenField := zap.Int64("results.tekton.dev/time-taken-ms", time.Since(startTime).Milliseconds())
```
adambkaplan, line 76: Potential feature - try and capture this in Prometheus metrics. We should discuss what sort of metrics make sense and what level of detail we want to record data.

pkg/watcher/reconciler/dynamic/dynamic_test.go
```diff
@@ -205,10 +206,33 @@ func TestReconcile_TaskRun(t *testing.T) {
 		}
 
 		_, err := trclient.Get(ctx, taskrun.GetName(), metav1.GetOptions{})
-		if !errors.IsNotFound(err) {
+		if !apierrors.IsNotFound(err) {
 			t.Fatalf("wanted NotFound, got %v", err)
 		}
 	})
+
+	t.Run("delete failed runs", func(t *testing.T) {
```
adambkaplan, line 214: I think this needs one more test case for "keep running TaskRuns" (the `Unknown` status.)

adambkaplan (changes requested):
I think we just need one more test case in the unit tests, otherwise lgtm.

pkg/watcher/reconciler/dynamic/dynamic.go
alan-ghelardi: I totally agree. Exposing relevant metrics from both API server and Watcher is an important step towards making the project production ready.

alan-ghelardi:
>I think this needs one more test case for "keep running TaskRuns" (the Unknown status.)

@adambkaplan , actually since the test is exercizing the deletion flow, if the object is running it should not be deleted at all. The test already covers this case by checking if the CompletionTime is unset and the grace period hasn't elapsed.

alan-ghelardi:
@adambkaplan  do you have any thoughts on the previous comment?
/assign @adambkaplan

adambkaplan:
/lgtm

>  if the object is running it should not be deleted at all. The test already covers this case by checking if the CompletionTime is unset and the grace period hasn't elapsed.

Makes sense to me. Looks good!

alan-ghelardi:
Weird, all checks passed, but the pr wasn't merged automatically. Wonder you need to approve manually in the GitHub UI because you had requested changes @adambkaplan ?

khrm (commented):
/kind bug

Issue #286, opened 2022-12-19
Allow users to pass label selectors to determine if a resource is eligible for deletion


alan-ghelardi:
### Feature request

Currently, the Watcher deletes a resource when the `completed_run_grace_period` flag is set and the time elapsed since the resource's completion matches the specified value. However, sometimes, users may set  up other tools to watch resources such as PipelineRuns and TaskRuns to perform certain actions. In those cases, it's  useful to wait until those tools complete their job before deleting the resources.

The Watcher could receive a new flag (say, `wait_for_labels`) to set a  list of label selectors to be matched against PipelineRuns and TaskRuns to determine whether the resource in question is eligible to be deleted. External tools that need to perform actions during the lifecycle of those resources must mark them with custom labels to signal to the Watcher that they have completed their job and the resources are eligible for deletion.

### Use case

This feature provides a flexible and decoupled way to external tools to communicate with the Results Watcher in order to synchronize their jobs. For instance, in our company, we've got a controller that stores completed runs in  our data lake. In order  to synchronize the actions performed by this controller and the Results watcher,  our controller would mark the stored runs with a custom label. The Results Watcher would be configured to wait until this label is present to delete the resources. The same principle is applicable to other integrations that can be developed eventually.

Besides of that, this feature would give people a more fine-grained control over deleted runs. If for some reason, users want to preserve certain PipelineRuns in the Kubernetes cluster even after archiving them in the Results API server, they just need to specify label selectors that match those runs.

adambkaplan:
@alan-ghelardi would a [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) work better for your situation? That lets controllers do a few things:

1. Indicate an object should not be deleted because some other work is being done.
2. Allow the controller to do any clean up work when an object needs to be deleted.

That said, I can understand why a finalizer would not be suitable as it is very easy for finalizers to be "orphaned" (ex: controller is stopped or upgraded to no longer use a finalizer). This can block cleanup of cluster resources.

alan-ghelardi:
@adambkaplan  usually I prefer to let the controller that owns the resource to deal with finalizers and the approach based on selectors looks like less intrusive and causes less friction in scenarios in which an incident occurs. For instance, if something causes tens of thousands of objects to get accumulated in the cluster and an administrator needs to delete them manually, they need to remove the finalizers to force the objects to be deleted.
[...]

PR #295, opened 2022-12-23
Allow users to set a label selector to filter objects to be deleted

alan-ghelardi:
This pull request resolves https://github.com/tektoncd/results/issues/286.

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -29,6 +29,7 @@ import (
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/controller"
@@ -122,6 +123,7 @@ func (r *Reconciler) addResultsAnnotations(ctx context.Context, o results.Object
 // than 0).
 // * The object is done and it isn't owned by other object.
 // * The configured grace period has elapsed since the object's completion.
+// * The object satisfies all label requirements defined in the supplied config.
 func (r *Reconciler) deleteUponCompletion(ctx context.Context, o results.Object) error {
 	logger := logging.FromContext(ctx)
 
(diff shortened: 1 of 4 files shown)


dibyom:
Nice!
/approve

dibyom:
@adambkaplan - one thing I just realized, we don't really have a PR template for this repo unlike others. I think we should add some. For this PR for instance, it would be useful to have a Release note

adambkaplan:
@dibyom :+1: on having a PR template and checks for release notes. I think for v0.5.0 we may have to hand-craft release notes.

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -156,6 +158,12 @@ func (r *Reconciler) deleteUponCompletion(ctx context.Context, o results.Object)
 		return controller.NewRequeueAfter(requeueAfter)
 	}
 
+	// Verify whether this object matches the provided label selectors
+	if selectors := r.cfg.GetLabelSelector(); !selectors.Matches(labels.Set(o.GetLabels())) {
+		logger.Debugw("Object doesn't match the required label selectors - requeuing to process later", zap.String("results.tekton.dev/label-selectors", selectors.String()))
+		return controller.NewRequeueAfter(1 * time.Minute)
```
adambkaplan, line 164: One minute is pretty fast for a re-queue of a completed PipelineRun/TaskRun, especially for something that likely won't get deleted. 10 minutes is a typical re-list interval for k8s controllers, I would go with that:

```suggestion
		return controller.NewRequeueAfter(10 * time.Minute)
```

adambkaplan (commented):
One bit about the re-queue interval, otherwise looks good to me.

pkg/watcher/reconciler/dynamic/dynamic.go
alan-ghelardi: What do you think about making this value configurable? Then users may decide what fits in their needs. I'm bringing it up because in our case at Nubank, we prefer to collect the PipelineRun/TaskRun as soon as possible because during high peaks, the cluster gets overloaded very quickly with thousands of simultaneous runs.
So, we could add a new flag (requeue_interval) whose default value would be 10m as you've suggested. For more specific scenarios, users may increase or decrease this value.

alan-ghelardi: Done @adambkaplan. Do you mind to have a second look?

avinal (approved):
Looks good to me.

khrm (commented):
/kind feature

Issue #296, opened 2022-12-26
Insertion of TaskRuns owned by a PipelineRun is broken


alan-ghelardi:
# Expected Behavior

All TaskRuns associated to a PipelineRun must be inserted properly into the Results API server.

# Actual Behavior

The insertion of child objects is unstable; some of the TaskRuns owned by a PipelineRun aren't inserted into the Results API.

# Steps to Reproduce the Problem

1. Create a pipeline made up of a few tasks.
2.  Create a few runs of this pipeline.
3. Inspect the database: some of the TaskRuns may not be recorded as expected.
4. Inspect the TaskRuns in the cluster: the missing TaskRuns in the database will contain the `results.tekton.dev/record` annotation with the same value as the same annotation of the parent PipelineRun.
5. Inspect the logs of the Watcher: the error `rpc error: code = FailedPrecondition desc = the etag mismatches` appears quite often while reconciling the PipelineRun and the missing TaskRuns in the database.    

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

Irrelevant

The root cause seems to be a race condition in the addition of the `results.tekton.dev/record` annotation. The Watcher patches the PipelineRun and the TaskRuns adding this annotation to set the corresponding record name. Underlying Watcher components read this annotation to determine the record name to be upserted into the API. Since the Tekton Pipelines controller propagates annotations from PipelineRuns to underlying TaskRuns, the Watcher may end up attempting to upsert the PipelineRun with the content of one of its child TaskRuns, what explains the aforementioned RPC error. At the end of the reconciliation, the PipelineRun gets updated again into the database and one or more of the conflicting TaskRuns will not be present in the API.

alan-ghelardi:
I'm looking into the issue and will open a pull request with a fix as soon as possible.

PR #298, opened 2022-12-29
Fix insertion of child resources

alan-ghelardi:
This pull request resolves https://github.com/tektoncd/results/issues/296.

From now on the Watcher reads the record name from the
`results.tekton.dev/record` annotation only if the object in question is a
top-level resource, i.e. isn't owned by other objects. If the object is a
TaskRun associated to a PipelineRun, the `results.tekton.dev/record` annotation
is propagated from the PipelineRun to the TaskRun what causes conflicts:
ultimately, both controllers end up competing to update the same record. This
causes many precondition failed errors slowing down the reconciliation and
causing incongruities: one or more TaskRuns aren't inserted in the database
because those records are overwritten by the PipelineRun's record.

In addition, I included debug log messages to facilitate the investigation of
the Watcher's behavior.

A follow-up is creating E2E tests to prevent future regressions in this
behavior. I'd like to treat it in a future pull request.

### DIFF
--- a/pkg/watcher/results/results.go
+++ b/pkg/watcher/results/results.go
@@ -24,6 +24,7 @@ import (
 	"github.com/tektoncd/results/pkg/watcher/convert"
 	"github.com/tektoncd/results/pkg/watcher/reconciler/annotation"
 	pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
+	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -32,6 +33,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"knative.dev/pkg/apis"
+	"knative.dev/pkg/logging"
 )
 
 // Client is a wrapper around a Results client that provides helpful utilities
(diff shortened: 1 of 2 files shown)


dibyom:
> If the object is a TaskRun associated to a PipelineRun, the results.tekton.dev/record annotation
is propagated from the PipelineRun to the TaskRun what causes conflicts

Ahh...this is an interesting edge case arising from our choice of propagating annotations from parent objects to child objects

dibyom (approved):
Thanks @alan-ghelardi - this looks good to me.

pkg/watcher/results/results.go
```diff
@@ -181,9 +192,16 @@ func resultName(o metav1.Object) string {
...
-	name, ok := o.GetAnnotations()[annotation.Record]
-	if ok {
-		return name
+	// Attempt to read the record name from annotations only if the object
+	// in question is a top-level record (i.e. it isn't owned by another
+	// object). Otherwise, the annotation containing the record name maybe
+	// was propagated by the owner what causes conflicts while upserting the
+	// object into the API. For further details, please see
+	// https://github.com/tektoncd/results/issues/296.
+	if isTopLevelRecord(o) {
```
adambkaplan, line 201: :+1: on the descriptive comment.

adambkaplan (approved):
/lgtm

khrm (commented):
/kind bug

PR #315, opened 2023-01-11
Fix DEVELOPMENT documentation

enarha:
Fix `create token` command, works with latest version of k8s.
Fix logging levels list.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -82,8 +82,11 @@ $ grpc_cli ls --channel_creds_type=ssl --ssl_target=tekton-results-api-service.t
 grpc.reflection.v1alpha.ServerReflection
 tekton.results.v1alpha2.Results
 
+# Create gRPC access token
+$ export GRPC_ACCESS_TOKEN=$(kubectl create token tekton-results-debug -n tekton-pipelines)
+
 # Makes a request to the Results service.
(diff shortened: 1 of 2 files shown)


docs/DEVELOPMENT.md
```diff
@@ -83,7 +83,7 @@ grpc.reflection.v1alpha.ServerReflection
 tekton.results.v1alpha2.Results
 
 # Makes a request to the Results service.
-$ grpc_cli call --channel_creds_type=ssl --ssl_target=tekton-results-api-service.tekton-pipelines.svc.cluster.local --call_creds=access_token=$(kubectl get secrets -n tekton-pipelines -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='tekton-results-debug')].data.token}"|base64 --decode) localhost:50051 tekton.results.v1alpha2.Results.ListResults 'parent: "default"'
+$ grpc_cli call --channel_creds_type=ssl --ssl_target=tekton-results-api-service.tekton-pipelines.svc.cluster.local --call_creds=access_token=$(kubectl create token tekton-results-debug -n tekton-pipelines) localhost:50051 tekton.results.v1alpha2.Results.ListResults 'parent: "default"'
```
avinal, line 86: I would suggest breaking this command into two, since the token is gonna be used multiple times, it would be nicer to save it in a variable and then reference it. WDYT?

`export SOME_TOKEN_NAME=$(kubectl create token tekton-results-debug -n tekton-pipelines)`

avinal (changes requested):
Nice work 😸

docs/DEVELOPMENT.md
enarha: I see this more as a demo code, but if people find it useful then why not :-). Updated.

enarha:
I found that the API documentation was listing `8080` as the port Prometheus is listening on. Changed to `9090` which is the port actually used.

khrm:
/approve

Looks good to me.

khrm (approved):

avinal (approved):

adambkaplan (approved):
/approve

/lgtm

khrm (commented):
/kind documentation

PR #316, opened 2023-01-12
Remove watcher reconciler unittest

enarha:
The test has race condition and the same functionality if covered by an e2e test.


khrm (commented):
/ok-to-test

adambkaplan (approved):
/approve

khrm (commented):
/kind cleanup

PR #317, opened 2023-01-12
Change the ClusterRoles for Logging

khrm:
Added `logs` as resources in ClusterRole readonly, readwrite and admin.


adambkaplan (approved):
/approve

khrm (commented):
/kind bug

Issue #318, opened 2023-01-13
REST access for Logs RPCs need to be accessed by GetRecord input pattern


sayan-biswas:
https://github.com/tektoncd/results/blob/357145ba7395933d9fbe9a5c3534db1d281c2885/proto/v1alpha2/api.proto#L99

In the above code, the REST handler for the API `GetLog` will set the input `Name = <parent>/results/<result-id>/logs/<record-id>.` But the GRPC handler expects the parameter `Name` in `the format <parent>/results/<result-id>/records/<record-id>.`

PR #301, opened 2023-01-03
[TEP-0117] Tekton Results Logs

sayan-biswas:
# Description

[[TEP-0117] Tekton Results Logs](https://github.com/tektoncd/community/blob/main/teps/0117-tekton-results-logs.md)

This PR extends [Tekton Results](https://github.com/tektoncd/results) to support log persistence outside of the host Kubernetes cluster. With this feature results will be able to store `PipelineRun` and `TaskRun` logs in the following persistent locations:

- File (PVC)
- S3 (S3 compatible storage)

# Release Notes
```release-note
Kubernetes version: 1.22.x
Tekton Pipeline version: 0.42.0

Add support for storing PipelineRun and TaskRun logs in persistent storage backends.
Refer to the TEP-0117 for more details.
```

PR #319, opened 2023-01-13
API fix for REST access

sayan-biswas:
This PR addresses the following [issue-318](https://github.com/tektoncd/results/issues/318)

**Fix**
- `logs` static path fragment added to differentiate between `GetRecords` and `GetLogs` in the generated HTTP mux, keeping the input field `name` same.
- `ListLogs` request/response changed to print the whole log record (similar to `ListRecords`)

**Note** : Merge this before [PR-301](https://github.com/tektoncd/results/pull/301)


khrm (commented):
/ok-to-test

proto/v1alpha2/api.proto
```diff
@@ -96,14 +96,14 @@ service Results {
 service Logs {
   rpc GetLog(GetLogRequest) returns (stream Log) {
     option (google.api.http) = {
-      get: "/v1alpha2/{name=*/results/*/logs/*}"
+      get: "/v1alpha2/logs/{name=*/results/*/records/*}"
     };
```
khrm, line 100: Please add `results` prefix for Results API.

Also, won't it be better to have rest APIs like below: 
`/apis/results.tekton.dev/v1alpha2/logs`

sayan-biswas: Right.. I think this `/apis/results.tekton.dev/v1alpha2/logs` pattern is better. This follows the kube api access pattern. 
I'll change this for all the RPCs

avinal: +1

khrm (approved):

adambkaplan (approved):
/approve

@sayan-biswas can we currently test this end to end? Or do we need to put more effort in (example - get tkn-results in shape to use as a client).

khrm:
We can test this using CLI `grpc_cli` call. That's how I did it.

khrm:
We can test this using CLI grpc_cli call. That's how I tested it, @adambkaplan .

khrm:
Sorry. I thought you were talking about different pr. No, I didn't test this.

avinal (approved):
Hi @adambkaplan IIRC we don't have any test yet for testing the Rest APIs e2e, so we can only manually test it for now. Here are some test results from my test.
image


Log API implementation is due in #301 
image

sayan-biswas:
> /approve
> 
> @sayan-biswas can we currently test this end to end? Or do we need to put more effort in (example - get tkn-results in shape to use as a client).

As of now, we can test e2e for Results and Records APIs, with `curl`. For logs API we need the server and watcher changes merged.

adambkaplan (commented):
I'm actually going to propose that we use `parents` as the leading field in the API. This will make the Results API similar to k8s, where `namespaces` leads namespace-scoped paths in the REST API [1]. This will also eliminate the semi-stutter we have with the `results` resource type.

```
/apis/results.tekton.dev/v1alpha2/parents/{parent=*}`
```

@dibyom @alan-ghelardi what do you think?

[1] https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris

alan-ghelardi:
Sounds good to me.

sayan-biswas:
@adambkaplan The parent field stores `namespace` name. So can we call it namespace in the path?
For  results/records
`/apis/results.tekton.dev/v1alpha2/namespaces/{name=*/results/*}`
`/apis/results.tekton.dev/v1alpha2/namespaces/{name=*/results/*/records/*}`

For logs
`/apis/results.tekton.dev/v1alpha2/namespaces/{name=*/results/*/logs/*}`

adambkaplan:
> @adambkaplan The parent field stores `namespace` name. So can we call it namespace in the path? For results/records `/apis/results.tekton.dev/v1alpha2/namespaces/{name=*/results/*}` `/apis/results.tekton.dev/v1alpha2/namespaces/{name=*/results/*/records/*}`
> 
> For logs `/apis/results.tekton.dev/v1alpha2/namespaces/{name=*/results/*/logs/*}`

Conceptually Results has the idea of a "parent" which is independent of Kubernetes. It _can_ be, and often is, a k8s namespace. But I believe the design intent is to allow for "other things" to be a parent (example - a git commit sha). @dibyom  and @wlynch may be able to shed more light here. Therefore I'm reluctant to tie the API completely to k8s REST conventions.

In practice - especially if RBAC-based auth is enabled - the parent is a namespace.

dibyom:
> Conceptually Results has the idea of a "parent" which is independent of Kubernetes. It can be, and often is, a k8s namespace. But I believe the design intent is to allow for "other things" to be a parent (example - a git commit sha). @dibyom and @wlynch may be able to shed more light here. Therefore I'm reluctant to tie the API completely to k8s REST conventions.

Yup! @adambkaplan is right - for instance if you were to share the same results API across multiple tekton clusters, the parent cloud be cluster/<cluster-id>/namespace/<ns>/results/<id>

https://github.com/tektoncd/community/blob/main/teps/0021-results-api.md#result-api has some more details!

sayan-biswas:
@adambkaplan @dibyom That makes sense, I'll go ahead with the `parents/{parent=*}` path.

adambkaplan:
Can confirm that the URL path is resolving as expected. The x509 error is an artifact of the e2e script not adding "localhost" as a CN:

```sh
$ curl --insecure https://localhost:8080/apis/results.tekton.dev/v1alpha2/parents/example-build/results
{"code":14, "message":"connection error: desc = \"trans    port: authentication handshake failed: x509: certificat    e is valid for tekton-results-api-service.tekton-pipeli    nes.svc.cluster.local, not localhost\"", "details":[]}
```

adambkaplan:
Hrm, something isn't right. When I try to curl the REST service (fixed separately), I consistently get "not found" responses:

```
$ kubectl get taskrun -n example-build -o yaml 
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  annotations:
    results.tekton.dev/record: example-build/results/6ae44060-0976-4796-be08-a3f4d8230eb3/records/6ae44060-0976-4796-be08-a3f4d8230eb3
    results.tekton.dev/result: example-build/results/6ae44060-0976-4796-be08-a3f4d8230eb3
...
$ curl -k https://localhost:8080/apis/results.tekton.dev/parents/example-build/results/6ae44060-0976-4796-be08-a3f4d8230eb3
{"code":5, "message":"Not Found", "details":[]}
```

sayan-biswas:
@adambkaplan. The api version `v1alpha2` is missing in the curl URI. Can you try with that once?

sayan-biswas:
Some results from my testing.. 

GetRecord
image

GetLog
image

khrm (commented):
/kind bug

PR #325, opened 2023-01-19
Pin Pipelines version for e2e testing

adambkaplan:
Pipelines v0.43.+ introduces the v1 APIs, which Results is not ready to support yet:

- Pin the version of Pipelines to v0.42.0 for e2e testing.
- Update README to indicate we don't support v0.43.0


dibyom:
> Pipelines v0.43.+ introduces the v1 APIs, which Results is not ready to support yet:

v1beta1 API should continue to work though.

adambkaplan:
I didn't document it fully - it seemed that when I was creating TaskRuns with v0.43, `kubectl get` would start returning the v1 representation. I suspect that we will still reconcile v1beta1 objects just fine. The real question is if we can reconcile `v1` objects (will we pick up their representation as `v1beta1`)?

I don't know enough about CRD versioning and the api-machinery to know the answer offhand.

adambkaplan:
With the logs items in the works, we may also be blocked on cli. See https://github.com/tektoncd/cli/issues/1837.

Another argument for moving the core logic behind `tkn [taskrun|pipelinerun] logs` out of of cli and to the pipelines repo.

adambkaplan:
/close

It looks like we can continue pulling the latest version of Pipelines while still reconciling against v1beta1 APIs.

PR #326, opened 2023-01-19
Install: add option for localhost DNS name

adambkaplan:
Update the install script to use localhost as an optional DNS alternative in the generated certificate. This can facilitate developer testing.

```release-note
NONE
```

### DIFF
--- a/test/e2e/01-install.sh
+++ b/test/e2e/01-install.sh
@@ -35,6 +35,13 @@ kubectl create secret generic tekton-results-postgres --namespace="tekton-pipeli
 echo "Generating TLS key pair..."
 set +e
 mkdir -p "${SSL_CERT_PATH}"
+
+SSL_INCLUDE_LOCALHOST=${SSL_INCLUDE_LOCALHOST:-"false"}
+altNames="DNS:tekton-results-api-service.tekton-pipelines.svc.cluster.local"
+if [ "$SSL_INCLUDE_LOCALHOST" = "true" ] ; then
+    altNames+=",DNS:localhost"
+fi
+
 openssl req -x509 \
         -newkey rsa:4096 \
         -keyout "${SSL_CERT_PATH}/tekton-results-key.pem" \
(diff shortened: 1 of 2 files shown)


alan-ghelardi:
Is  it worth documenting this option in the docs/development.md file?

adambkaplan:
Updated the doc of the e2e install script. Once #319 merges, I think we can update the debugging doc to use `curl` instead of the gRPC client. And if we fix up `tkn-results`, we can advocate that as the right way to go.

avinal (approved):

test/e2e/01-install.sh
```diff
@@ -66,7 +73,7 @@ set -e
 kubectl create secret tls -n tekton-pipelines tekton-results-tls --cert="${SSL_CERT_PATH}/tekton-results-cert.pem" --key="${SSL_CERT_PATH}/tekton-results-key.pem" || true
 
 echo "Installing Tekton Results..."
-kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply -f -
+kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply --sbom=none -f -
```
khrm, line 76: Why did we disable sbom?

avinal: For local testing, SBOM isn't useful. SBOM generation adds extra time.

adambkaplan: note that this is also for e2e testing. At times I have tested with quay.io as my image repo, and pushing the SBOM does not always succeed.

avinal: This was a known issue but it should be fixed in the latest version: https://github.com/ko-build/ko/issues/532. I haven't tried it recently.

adambkaplan: I may have to try and reproduce the issue. Sounds like an issue with quay.io.

adambkaplan: Looks like quay.io doesn't like `spdx+json` media types in some circumstances. Commented in https://github.com/ko-build/ko/issues/660#issuecomment-1412739253 - it looks like ko has support for `text/spdx+json` on the long term roadmap?

avinal:
@adambkaplan Please rebase.

avinal (approved):
Looks good to me

sayan-biswas (approved):

Issue #311, opened 2023-01-09
Make the database auto-migration optional on the API server's startup


alan-ghelardi:
### Feature request

Currently, the database's schema is described in the [results.sql](https://github.com/tektoncd/results/blob/main/schema/results.sql) file, which is installed upon the [database's startup](https://github.com/tektoncd/results/blob/main/config/201-sql-deployment.yaml) as well as in all required test cases. In addition, the database auto-migration is executed upon the [API server's startup](https://github.com/tektoncd/results/blob/main/pkg/api/server/v1alpha2/server.go#L56).

After [bumping](https://github.com/tektoncd/results/pull/310) gorm as well as their related dependencies, unit and integration tests got broken due to conflicts in these two approaches.

In part, the issue is related to the fact that the schema defined in the results.sql file and in the gorm models are incongruent. For instance, the records.data column is declared as jsonb in the DDL script while the data type is missing in the corresponding [field](https://github.com/tektoncd/results/blob/main/pkg/api/server/db/model.go#L69). This causes the records.data column to be defined as bytea in the Postgres database.

## Proposal

- Fix the gorm models to generate compatible schemas with the DDL script provided by the project.
- Make the database auto-migration an opt-in. Then, users may decide to enable the database auto-migration in their installations or keep it disabled and use the DDL script bundled into the provided kustomization. This might be useful to allow users to customize the DDL file, for instance, to create additional indexes.

##  Related pull requests

- https://github.com/tektoncd/results/pull/310.
- https://github.com/tektoncd/results/pull/290.

PR #310, opened 2023-01-09
Bump Gorm and related dependencies

alan-ghelardi:
Resolves https://github.com/tektoncd/results/issues/273.

##  What

This pull request is a cherry-pick of
https://github.com/tektoncd/results/pull/300, including the bump of the
following gorm dependencies to fix compilation errors:

-       gorm.io/driver/postgres to v1.4.6.
- gorm.io/driver/sqlite to v1.4.4.
- github.com/mattn/go-sqlite3 to v1.14.16.

See the comment added by dependabot on the gorm.io/gorm's bump:

Bumps [gorm.io/gorm](https://github.com/go-gorm/gorm) from 1.23.4 to 1.24.3.
- [Release notes](https://github.com/go-gorm/gorm/releases)
- [Commits](https://github.com/go-gorm/gorm/compare/v1.23.4...v1.24.3)

---
cccupdated-dependencies:
- dependency-name: gorm.io/gorm
  dependency-type: direct:production
  update-type: version-update:semver-minor
...

## Issues

This bump reveals an issue with the database auto-migration performed on the API
server's startup, which is conflicting with the schema installed by the
results.sql file. We cannot merge this pull request until this issue isn't
resolved. I'm going to open a separate issue to track this problem.

PR #334, opened 2023-01-27
Upgrade Gorm models and make database auto-migration optional

alan-ghelardi:
Resolves https://github.com/tektoncd/results/issues/311 and
https://github.com/tektoncd/results/issues/289.

## What

- Sync Gorm models with the most recent schema (previously installed via the
sql-initdb ConfigMap).
- Delete the sql-initdb ConfigMap (alter the Postgres StatefulSet to stop
installing the schema on startup).
- Make the database auto-migration an opt-out. It is enabled by default because
the Gorm auto-migration is safe (it doesn't attempt to remove unused
fields/tables), but users may disable it via config and install a custom schema
or manage migrations by themselves.

## Why

As of this pull request we no longer need to keep models and the DDL script
aligned, in fact, they weren't. Any changes may be done directly in Gorm models
and they'll be reflected automatically in the database.


dibyom (commented):
This looks good to me !

Thanks @alan-ghelardi

dibyom (approved):

alan-ghelardi:
Reminder: after merging this pr, we can rebase and merge https://github.com/tektoncd/results/pull/310 to bump Gorm and related dependencies.

adambkaplan:
Looks good to me - please squash your commits and I'll add the lgtm.

adambkaplan:
Note - if we find that size checks are causing things to break, we can always increase them with zero end user impact. We just can't go the other way and reduce column sizes.

adambkaplan (approved):
/lgtm

khrm (commented):
/kind cleanup

PR #335, opened 2023-01-30
Fixing the install script for mac M1

prietyc123:
This changes aims to fix the install script for running on mac M1. Workaround for  [issue](https://github.com/ko-build/ko/issues/946)

Signed-off-by: Priti Kumari [pkumari@redhat.com](mailto:pkumari@redhat.com)

### DIFF
--- a/test/e2e/01-install.sh
+++ b/test/e2e/01-install.sh
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# shellcheck disable=SC2181 # To ignore long command exit code check
+
 set -e
 
 export KO_DOCKER_REPO=${KO_DOCKER_REPO:-"kind.local"}
@@ -24,11 +26,11 @@ ROOT="$(git rev-parse --show-toplevel)"
 
 echo "Installing Tekton Pipelines..."
 TEKTON_PIPELINE_CONFIG=${TEKTON_PIPELINE_CONFIG:-"https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml"}
-kubectl apply --filename ${TEKTON_PIPELINE_CONFIG}
+kubectl apply --filename "${TEKTON_PIPELINE_CONFIG}"
 
 echo "Generating DB secret..."
 # Don't fail if the secret isn't created - this can happen if the secret already exists.
-kubectl create secret generic tekton-results-postgres --namespace="tekton-pipelines" --from-literal=POSTGRES_USER=postgres --from-literal=POSTGRES_PASSWORD=$(openssl rand -base64 20) || true
+kubectl create secret generic tekton-results-postgres --namespace="tekton-pipelines" --from-literal=POSTGRES_USER=postgres --from-literal=POSTGRES_PASSWORD="$(openssl rand -base64 20)" || true
 
 echo "Generating TLS key pair..."
 set +e
@@ -66,13 +68,14 @@ set -e
 kubectl create secret tls -n tekton-pipelines tekton-results-tls --cert="${SSL_CERT_PATH}/tekton-results-cert.pem" --key="${SSL_CERT_PATH}/tekton-results-key.pem" || true
 
 echo "Installing Tekton Results..."
-kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply -f -
+extra_ko_params="linux/$(go env GOARCH)"
+kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply --platform="$extra_ko_params" -f -
 
 echo "Fetching access tokens..."
-mkdir -p ${SA_TOKEN_PATH}
+mkdir -p "${SA_TOKEN_PATH}"
 service_accounts=(all-namespaces-read-access single-namespace-read-access)
 for service_account in "${service_accounts[@]}"; do
-    kubectl create token $service_account > ${SA_TOKEN_PATH}/$service_account
+    kubectl create token "$service_account" > "${SA_TOKEN_PATH}"/"$service_account"
     echo "Created ${SA_TOKEN_PATH}/$service_account"
 done
 


test/e2e/01-install.sh
```diff
@@ -66,13 +66,17 @@ set -e
 kubectl create secret tls -n tekton-pipelines tekton-results-tls --cert="${SSL_CERT_PATH}/tekton-results-cert.pem" --key="${SSL_CERT_PATH}/tekton-results-key.pem" || true
 
 echo "Installing Tekton Results..."
-kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply -f -
+if [[ "$(uname -a)" == *ARM64* && "$(uname -m)" == x86_64* ]]; then
+    kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply --platform=linux/arm64 -f -
```
prietyc123, line 70: Do we need to change this directory to `${ROOT}/config` ?

test/e2e/01-install.sh
```diff
@@ -66,13 +66,17 @@ set -e
...
+else
+    kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply -f -
```
prietyc123, line 72: same as [comment](https://github.com/tektoncd/results/pull/335/files#r1090231937)

test/e2e/01-install.sh
enarha: `${ROOT}/config` is included https://github.com/tektoncd/results/blob/main/test/e2e/kustomize/kustomization.yaml#L16 . Does that answer your question?

test/e2e/01-install.sh
```diff
@@ -66,13 +68,17 @@ set -e
 kubectl create secret tls -n tekton-pipelines tekton-results-tls --cert="${SSL_CERT_PATH}/tekton-results-cert.pem" --key="${SSL_CERT_PATH}/tekton-results-key.pem" || true
 
 echo "Installing Tekton Results..."
-kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply -f -
+if [[ "$(uname -a)" == *ARM64* && "$(uname -m)" == x86_64* ]]; then
```
enarha, line 71: For the first match I'll suggest using Bash's regex match operator `=~` which is better suited for this type of check, so `"$(uname -a)" == *ARM64*` could become `"$(uname -a)" =~ ARM64`.

In the second part, why do you need the `*` in `x86_64*`? If `"$(uname -m)" == "x86_64"` does not match, I'll suggest again using the `=~` operator.

Instead of repeating the `kubectl kustomize ...` command in the if and else clauses, you can do:
`[[ "$(uname -a)" == *ARM64* && "$(uname -m)" == x86_64* ]] && extra_ko_params="--platform=linux/arm64"`
and then
`kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply $extra_ko_params -f -`. If `extra_ko_params` is not set, it'll return empty string which won't affect the `ko` command.

prietyc123: Done

prietyc123: @enarha empty string throws error
```
error: Unexpected args: []
See 'kubectl apply -h' for help and examples
```

test/e2e/01-install.sh
```diff
@@ -66,13 +68,17 @@ set -e
 kubectl create secret tls -n tekton-pipelines tekton-results-tls --cert="${SSL_CERT_PATH}/tekton-results-cert.pem" --key="${SSL_CERT_PATH}/tekton-results-key.pem" || true
 
 echo "Installing Tekton Results..."
-kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply -f -
+if [[ "$(uname -a)" =~ "ARM64" && "$(uname -m)" =~ "x86_64" ]]; then
+    extra_ko_params="linux/arm64"
+fi
```
adambkaplan, line 73: I don't see `extra_ko_params` being initialized outside of this if block, leading to potential problems.

Does it not make sense to use the go OS/ARCH env values?

```suggestion
exra_ko_params="$(go env GOOS)/$(go env GOARCH)"
```

adambkaplan (commented):
/approve

Indicating that this PR is providing a fix to a known issue, so that other community members can lgtm once code suggestions are addressed.

test/e2e/01-install.sh
enarha: The problem is you've put quotes around `$extra_ko_params`. Without them the proposed code works.

```[ results] -> kubectl kustomize "./test/e2e/kustomize" | ko apply "$extra_ko_params" -f -
error: Unexpected args: []
See 'kubectl apply -h' for help and examples```

```[ results] -> kubectl kustomize "./test/e2e/kustomize" | ko apply $extra_ko_params -f -
2023/01/30 20:38:51 Using base distroless.dev/static:latest@sha256:01dfcc8d455ddb111f496eca6e5fe88aa5efac96a7f08b269f1a259eb5a2f5fc for github.com/tektoncd/results/cmd/api
...```

Anyway, Adam's suggestion to always set the platform is a better approach.

File: test/e2e/01-install.sh (hunk 4 above)
Comment (enarha) [reply test/e2e/01-install.sh:73]: Agree, it does.

File: test/e2e/01-install.sh [hunk 5]
```diff
@@ -66,13 +68,17 @@ set -e
 kubectl create secret tls -n tekton-pipelines tekton-results-tls --cert="${SSL_CERT_PATH}/tekton-results-cert.pem" --key="${SSL_CERT_PATH}/tekton-results-key.pem" || true
 
 echo "Installing Tekton Results..."
-kubectl kustomize "${ROOT}/test/e2e/kustomize" | ko apply -f -
+if [[ "$(uname -a)" =~ "ARM64" && "$(uname -m)" =~ "x86_64" ]]; then
```
enarha, line 71: As suggested that check could go away, but for the record when using the `=~` operator, the right-hand side should be without quotes. See an example:
```
[ ~] -> [[ "abc" =~ "ab." ]] && echo "true"
[ ~] -> [[ "abc" =~ ab. ]] && echo "true"
true
```
The problem is that in quotes `.` loses it's special meaning to match a single character and instead is treated as `.` character part of the string.

test/e2e/01-install.sh
prietyc123: > Does it not make sense to use the go OS/ARCH env values?

As ko only builds base images in format
image
I think using `exra_ko_params="linux/$(go env GOARCH)"` would make sense. WDYT? Because on mac it will give `Darwin/arm64`

test/e2e/01-install.sh
prietyc123: > The problem is you've put quotes around $extra_ko_params

ohh damn, I put quotes to avoid shellcheck error. But yes I think always putting `extra_ko_params` should work

test/e2e/01-install.sh
prietyc123: Done address the changes. PTAL 🙂

enarha: /lgtm

avinal (approved):
/lgtm

khrm (commented):
/kind misc

Issue #336, opened 2023-01-31
The API server causes almost 100% CPU consumption on Amazon RDS 


alan-ghelardi:
## Context

We're running Tekton Results on Amazon EKS using a RDS instance as the database. Our CI/CD cluster hosts hundreds of thousands of runs daily and after 2 months  deployed to production, the Results database contains almost 2 millions of rows in the records table.

During high peaks we're observing that the CPU consumption of the RDS instance stands around 99,80% without significant oscilations.

## Debugging and root cause

We've troubleshooted the database and didn't find anything abnormal (e.g. no leaking connections, no idle transactions, etc.). After looking at certain performance insights, however, we've found out that the following query (which runs all the time) was consuming almost all database resources:

`SELECT * FROM "records" WHERE "records"."name" = $1 ORDER BY "records"."id" LIMIT ?`

This query is executed to find a given record and for some reason, the API server isn't using all columns that make up the records_by_name index to optimize the select.

The root cause seems to be a mistake at this [line](https://github.com/tektoncd/results/blob/main/pkg/api/server/v1alpha2/records.goL#128): instead of setting the Parent and ResultName fields in the Record object, we're passing those values to the Record.Result field what prevents Gorm from leveraging the existing index for the query.

alan-ghelardi:
@marcodalcin

PR #338, opened 2023-01-31
Fix the select statement to leverage the existing records_by_name index

alan-ghelardi:
Resolves https://github.com/tektoncd/results/issues/336 which contains further
details on the rationale.

This is a low-hanging fruit, but causes a huge impact in terms of performance at
scale. After this fix, the GetRecord operation started to use the indexed
columns to find records and we reduced the CPU utilization of our Postgres
instance from almost 100% to around 3% during high peaks.

### DIFF
--- a/pkg/api/server/v1alpha2/records.go
+++ b/pkg/api/server/v1alpha2/records.go
@@ -123,9 +123,14 @@ func (s *Server) GetRecord(ctx context.Context, req *pb.GetRecordRequest) (*pb.R
 }
 
 func getRecord(txn *gorm.DB, parent, result, name string) (*db.Record, error) {
+	// Note: set the Parent, ResultName and Name fields in the model used to
+	// query the database to take advantage of the records_by_name composite
+	// index. Although the Name is an unique value as well, leveraging the
+	// index speeds up the query significantly. See
+	// https://github.com/tektoncd/results/issues/336.
 	store := &db.Record{}
 	q := txn.
-		Where(&db.Record{Result: db.Result{Parent: parent, Name: result}, Name: name}).
+		Where(&db.Record{Parent: parent, ResultName: result, Name: name}).
 		First(store)
 	if err := errors.Wrap(q.Error); err != nil {
 		return nil, err


alan-ghelardi:
Cudos for @marcodalcin who helped me in the investigation.

dibyom:
Thanks @alan-ghelardi and @marcodalcin

/approve

alan-ghelardi:
assign @adambkaplan

alan-ghelardi:
H/assign @adambkaplan

pkg/api/server/v1alpha2/records.go
```diff
@@ -125,7 +125,7 @@ func (s *Server) GetRecord(ctx context.Context, req *pb.GetRecordRequest) (*pb.R
 func getRecord(txn *gorm.DB, parent, result, name string) (*db.Record, error) {
 	store := &db.Record{}
 	q := txn.
-		Where(&db.Record{Result: db.Result{Parent: parent, Name: result}, Name: name}).
+		Where(&db.Record{Parent: parent, ResultName: result, Name: name}).
```
adambkaplan, line 128: Please add a comment here explaining why this is using the `ResultName` field  (if I'm not mistaken, it's our indexed foreign key to the Results table?)

It took me a moment to figure out that the `db.Result` object created an unnecessary join in the query :sob:

alan-ghelardi: Done @adambkaplan. Just for clarifying if anyone else is reading this thread eventually, there's a composite index (parent, result_name, name) in the records table. The name is an unique value as well (that's why the query was working despite of this issue), but it causes a full database scan if used alone. 
Afaik, Gorm doesn't run a join implicitly to load the object that corresponds to a foreign key, but anyhow, in that query populating the Record.Result field has no effect because we want to find the record that matches the provided parent, resultName and name and not load the association.

adambkaplan (approved):
/lgtm

:tada:

khrm (commented):
/kind cleanup

PR #339, opened 2023-02-01
Do not create the parent result if the object being reconciled is a child resource

alan-ghelardi:
## Context

Currently, the watcher attempts to create the parent result whenever it is
missing regardless if the object being reconciled is a top-level resource,
that's to say, isn't owned by another resource. In highly concurrent scenarios
this approach leads to race conditions in which various TaskRuns being
reconciled concomitantly with their PipelineRun ends up attempting to create the
same result almost simultaneously. This causes lots of unnecessary requests
being made to the API server which result in unique constraint violations being
raised by the database. Even though the unique constraint enforced by the
database prevents the duplication of records, the Watcher ends up raising
unnecessary errors cluttering the logs.

## Proposal

This pull request alters the Watcher to create the parent result only when the
object being reconciled is a top-level one. If the result client is processing a
child object and the result doesn't exist yet, it returns a specific error that
is interpreted by the controller as a signal to requeue the object in
question. I believe that this approach made the code simpler and eliminate a
bunch of unique constraint violations that show up very often in highly
concurrent environments. While analyzing the logs emitted by the Watcher, those
errors catch the attention unnecessarily and may cause confusion in end users
who aren't so familiar with the project internals.

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	"errors"
+
 	"github.com/jonboulle/clockwork"
 	pipelinev1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/results/pkg/watcher/convert"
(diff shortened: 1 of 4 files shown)


adambkaplan:
bump @alan-ghelardi  - pr needs rebase.

Is there still a root issue here? Did #298 fix this?

adambkaplan:
@alan-ghelardi is this still an issue that warrants a fix?

PR #342, opened 2023-02-02
Fix go generate: Protobuf generation

khrm:
Put `.` in place of `./proto` from command. go generate doesn't work in Fedora 37(go 1.19.5).

### DIFF
--- a/proto/README.md
+++ b/proto/README.md
@@ -4,6 +4,5 @@
 2. Rebuild the generated Go code
 
 ```sh
-$ cd v1alpha2
-$ go generate ./proto/
+$ go generate -v ./...
 ```


khrm (commented):
/kind bug

vdemeester (approved):

proto/README.md
```diff
@@ -4,6 +4,5 @@
 2. Rebuild the generated Go code
 
 ```sh
-$ cd v1alpha2
-$ go generate ./proto/
```
enarha, line 8: The problem seems to be related to wrong directory name. If you `cd` to `v1alpha2`, there is no `proto` directory there. Your fix is OK, another option is to run `go generate ./v1alpha2/` from the `proto` directory.

khrm: We will also need to generate `pipeline` directory. So I wrote `go generate ./v1alpha2/`

enarha: I'm a bit confused. The change suggest running `go generate -v ./...`.
I think it also helps specifying which is the directory the command in run from. If called from the repo root, more changes are generated compared to when running from {root}/proto.

khrm: Sorry, I meant I wrote `go generate -v ./...` because we want to run generate for both directories.  Since this Readme is in the proto directory, it implies running from this directory.

enarha: OK.
/lgtm

PR #373, opened 2023-02-28
Feature: Add impersonation to tekton results API

### LINKED ISSUE #309: User Impersonation for RBAC
### Feature request

Support Kubernetes [user impersonation ](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation) when authenticating and authorizing requests.

### Use case

1. Run a proxy in front of Results to access data across multiple clusters, which can forward requests to one of many clusters running Tekton Results.
3. Support `tkn results --as` flag, in a similar fashion as `kubectl <command> --as`.

### Implementation Details

The Subject Access Review API supports all fields that should be provided for impersonation.
The main work Results needs to do is:

1. Extract the impersonation headers from the incoming request.
2. Do appropriate Subject Access Review(s) if impersonation headers are present.
3. If the incoming user is allowed to do impersonation, run the main SAR check as the impersonated user.

sayan-biswas:
# Changes

Resolves #309 

This feature adds Kubernetes impersonation header processing, impersonation access check for requester and RBAC for tekton results resources with impersonated user data. 

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation

# Submitter Checklist

# Release Notes

```release-note
Kubernetes impersonation headers can be used when authorizing access to Tekton Results APis.

ACTION REQUIRED: The NO_AUTH config parameter was renamed to AUTH_DISABLE
```


pkg/api/server/v1alpha2/auth/impersonation/impersonation.go
```diff
@@ -0,0 +1,217 @@
...
+					Subresource: resourceAttribute.Subresource,
+					Verb:        resourceAttribute.Verb,
+				},
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			return err
+		}
+		if !sar.Status.Allowed {
+			return fmt.Errorf("forbidden: '%s' doesn't have access to impersonate '%s'", requester, resourceAttribute.Resource)
```
adambkaplan, line 190: small change to error message:

```suggestion
			return fmt.Errorf("forbidden: '%s' doesn't have permission to impersonate '%s'", requester, resourceAttribute.Resource)
```

pkg/api/server/v1alpha2/auth/rbac.go
```diff
@@ -89,11 +109,32 @@ func (r *RBAC) Check(ctx context.Context, namespace, resource, verb string) erro
 			continue
 		}
 
+		user := tr.Status.User.Username
+		UID := tr.Status.User.UID
+		groups := []string{"tekton.dev"}
+		extra := map[string]authzv1.ExtraValue{}
+
+		// Check whether the authenticated user has access to impersonate
```
adambkaplan, line 117: Replace "access" with permission. Admins grant _access_, users receive _permission_.

```suggestion
		// Check whether the authenticated user has permission to impersonate
```

pkg/api/server/v1alpha2/auth/rbac.go
```diff
@@ -38,22 +39,41 @@ import (
 // This checks RBAC permissions in the `results.tekton.dev` group, and assumes
 // checks are done at the namespace
 type RBAC struct {
-	authn authnclient.AuthenticationV1Interface
-	authz authzclient.AuthorizationV1Interface
+	impersonation bool
```
adambkaplan, line 42: Maybe call this `allowImpersonation` to avoid confusion.

pkg/api/server/v1alpha2/auth/rbac.go
```diff
@@ -114,3 +155,19 @@ func (r *RBAC) Check(ctx context.Context, namespace, resource, verb string) erro
...
+func convertExtra(extra map[string][]string) map[string]authzv1.ExtraValue {
+	var newExtra map[string]authzv1.ExtraValue
+	for key, value := range extra {
+		newExtra[key] = value
+	}
+	return newExtra
+}
+
+// WithImpersonation is an option function to enable Impersonation
+func WithImpersonation() Option {
```
adambkaplan, line 169: With this kind of builder/factory pattern, you should accept an argument so it can be explicitly set to true or false.

```suggestion
func WithImpersonation(enabled bool) Option {
```

pkg/api/server/v1alpha2/auth/rbac.go
```diff
@@ -38,22 +39,41 @@ import (
...
 }
 
 func (r *RBAC) Check(ctx context.Context, namespace, resource, verb string) error {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
 		return status.Error(codes.Unauthenticated, "unable to get context metadata")
 	}
+
+	// Parse Impersonation header if the feature is enabled
+	var i *impersonation.Impersonation
```
adambkaplan, line 67: Please use a more descriptive variable name here.

```suggestion
	var impersonator *impersonation.Impersonation
```

pkg/api/server/v1alpha2/auth/impersonation/impersonation.go
```diff
@@ -0,0 +1,217 @@
...
+	} else {
+		i.userInfo = nil
+
+		// Impersonate-User header is mandatory.
+		// https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation
+		if hasGroups || hasExtra || hasUID {
+			return ErrorImpersonateUserRequired
+		} else {
+			return ErrorNoImpersonationData
+		}
```
adambkaplan, line 153: Should we return an error here? This seems to imply that _all_ requests require impersonation, which is certainly not the case for requests that come in from the watcher.

config/base/env/config
```diff
@@ -10,7 +10,8 @@ REST_PORT=8080
 PROMETHEUS_PORT=9090
 TLS_HOSTNAME_OVERRIDE=
 TLS_PATH=/etc/tls
-NO_AUTH=false
+AUTH_DISABLE=false
```
adambkaplan, line 13: This absolutely needs to be called out as a breaking change (if we decide to accept it).

config/base/env/config
```diff
@@ -10,7 +10,8 @@ REST_PORT=8080
...
+AUTH_IMPERSONATE=false
```
adambkaplan, line 14: Personally, I think we should set the default config to `true` in our release. At minimum this will ensure that we run end to end tests with this enabled, reducing our test matrix.

cmd/api/main.go
```diff
@@ -186,7 +193,7 @@ func main() {
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	mux := runtime.NewServeMux()
+	mux := runtime.NewServeMux(runtime.WithIncomingHeaderMatcher(impersonation.HeaderMatcher))
```
adambkaplan, line 196: Should this be added if the impersonation check is disabled?

adambkaplan (commented):
/approve

Marking this as an approved feature to add. In the current form, however, we need to ensure that e2e tests work with this feature enabled. @dibyom do other projects have a mechanism to test features e2e where the feature flag is enabled _and_ disabled?

cmd/api/main.go
sayan-biswas: The default header matcher is implicit in `runtime.NewServeMux()`. This custom function is added so that the requester can send impersonation header with name `Impersonate-User` instead of GRPC gateway header format, `Grpc-Metadata-Impersonate-User`.

But I'll still put in this in a condition.

config/base/env/config
sayan-biswas: Default set to `true`.

pkg/api/server/v1alpha2/auth/rbac.go
sayan-biswas: updated

pkg/api/server/v1alpha2/auth/rbac.go
sayan-biswas: Added argument

pkg/api/server/v1alpha2/auth/rbac.go
sayan-biswas: renamed to `allowImpersonation `

pkg/api/server/v1alpha2/auth/rbac.go
sayan-biswas: :thumbsup: Changed.

pkg/api/server/v1alpha2/auth/impersonation/impersonation.go
sayan-biswas: :thumbsup: Changed

cmd/api/main.go
sayan-biswas: https://github.com/tektoncd/results/blob/06adf689eb14324be872c43787d6d68861df5af9/cmd/api/main.go#L109-L112

pkg/api/server/v1alpha2/auth/impersonation/impersonation.go
sayan-biswas: I returned error here so that the `NewImpersonation` function doesn't return an empty or nil `Impersonation` struct without returning any error.
This error will not stop the processing of the requests, because it's handled here.

https://github.com/tektoncd/results/blob/06adf689eb14324be872c43787d6d68861df5af9/pkg/api/server/v1alpha2/auth/rbac.go#L71-L74

The idea here is if no impersonation header is present we, continue processing with a nil `Impersonation` struct, which is checked later while performing impersonation check. But if any other impersonation header is present without having `Impersonate-User` header, then we stop processing of that request. This is the same behaviour in the Kubernetes impersonation filter implementation.

config/base/env/config
sayan-biswas: I think this name makes it easier to categorise the options in the `config`, prepend with a short type description for the option.

adambkaplan (commented):
Almost there! Code looks good, I added a suggestion to add a code comment.

Can you also please add a section to the Authorization doc that indicates we support Kubernetes impersonation headers [1]? You don't need to explain what user impersonation is - a link to the Kubernetes doc [2] is sufficient. Just focus on how to enable the feature and how Results clients (gRPC, REST) can use it.

[1] https://github.com/tektoncd/results/tree/main/docs/api#authenticationauthorization
[2] https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation

pkg/api/server/v1alpha2/auth/rbac.go
```diff
@@ -38,22 +39,41 @@ import (
...
+
+	// Parse Impersonation header if the feature is enabled
+	var impersonator *impersonation.Impersonation
+	var err error
+	if r.allowImpersonation {
+		impersonator, err = impersonation.NewImpersonation(md)
+		if err != nil && err != impersonation.ErrorNoImpersonationData {
+			log.Println(err)
+			return status.Error(codes.Unauthenticated, "invalid impersonation data")
+		}
```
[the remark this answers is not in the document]
sayan-biswas: Done

adambkaplan:
/assign @dibyom @alan-ghelardi 

There is a breaking configuration change here.

khrm:
Shouldn't we introduce a deprecation notice instead of directly breaking it? We can remove the `NO_AUTH` in the next release with the ACTION required notice. In this release, we can have a deprecation notice.

config/base/env/config
```diff
@@ -10,7 +10,8 @@ REST_PORT=8080
 PROMETHEUS_PORT=9090
 TLS_HOSTNAME_OVERRIDE=
 TLS_PATH=/etc/tls
-NO_AUTH=false
+AUTH_DISABLE=false
+AUTH_IMPERSONATE=true
```
dibyom, line 14: Is this enabled by default? Doc above states that it is disabled by default?

dibyom:
> Shouldn't we introduce a deprecation notice instead of directly breaking it? We can remove the NO_AUTH in the next release with the ACTION required notice. In this release, we can have a deprecation notice.

Technically, we are allowed to make breaking changes in alpha apis. Though generally what we have done in other projects is what @khrm is suggesting here

dibyom (approved):

config/base/env/config
sayan-biswas: Ohh.. I missed this. 
Updated doc.

sayan-biswas:
> Shouldn't we introduce a deprecation notice instead of directly breaking it? We can remove the `NO_AUTH` in the next release with the ACTION required notice. In this release, we can have a deprecation notice.

Agreed! But this flag is mostly for development, to bypass the RBAC processing.
Even if this flag is missing from the config, the parses will assume the value as false and continue. Which is the default behaviour anyways. 
Also the release will ship the config file which will anyways have the modified parameter.

sayan-biswas:
> Almost there! Code looks good, I added a suggestion to add a code comment.
> 
> Can you also please add a section to the Authorization doc that indicates we support Kubernetes impersonation headers [1]? You don't need to explain what user impersonation is - a link to the Kubernetes doc [2] is sufficient. Just focus on how to enable the feature and how Results clients (gRPC, REST) can use it.
> 
> [1] https://github.com/tektoncd/results/tree/main/docs/api#authenticationauthorization [2] https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation

Added code comment and updated doc.

adambkaplan:
@alan-ghelardi just want to check with you that renaming the `NO_AUTH` config setting to `AUTH_DISABLE` is acceptable (since this feature was your contribution).

Otherwise LGTM.

PR #405, opened 2023-03-18
Use same listener for  gRPC and REST request

### LINKED ISSUE #406: Allow user to send gRCP and REST traffic on same port
### Feature request

Allow user to send gRCP and REST traffic on same port

### Use case

Currently we are using two separate port for sending gRPC and REST requests. For users who to use both gRPC and REST client would require two endpoints to maintain. We should enhance the server to accept both type of requests on the same port.

sayan-biswas:
# Changes
Enables gRPC and REST traffic to be listened on the same port.
Creates only one HTTP listener to accept both gRPC and REST requests. A custom handler is used to identify gRPC request and sent it to the gRPC server wrapped in a h2c handler. Other requests are sent to the multiplexer created for REST endpoints. 

Resolves #406 

# Submitter Checklist

# Release Notes

```release-note
ACTION REQUIRED: Config parameters GRPC_PORT and REST_PORT are replaced by a single parameter SERVER_PORT, which handles both gRPC and REST requests.
```


cmd/api/main.go
```diff
@@ -187,28 +179,39 @@ func main() {
...
+		log.Fatal(http.ListenAndServe(":"+serverConfig.SERVER_PORT, grpcHandlerFunc(gs, mux)))
 	} else {
-		log.Fatal(http.ListenAndServeTLS(":"+serverConfig.REST_PORT, path.Join(serverConfig.TLS_PATH, "tls.crt"), path.Join(serverConfig.TLS_PATH, "tls.key"), mux))
+		log.Fatal(http.ListenAndServeTLS(":"+serverConfig.SERVER_PORT, path.Join(serverConfig.TLS_PATH, "tls.crt"), path.Join(serverConfig.TLS_PATH, "tls.key"), grpcHandlerFunc(gs, mux)))
 	}
 
 }
 
+// grpcHandlerFunc forwards the request to gRPC server based on the Content-Type header.
+func grpcHandlerFunc(grpcServer *grpc.Server, otherHandler http.Handler) http.Handler {
```
adambkaplan, line 205: nit - refer to the other as "rest" or "http"
```suggestion
func grpcHandlerFunc(grpcServer *grpc.Server, httpServer http.Handler) http.Handler {
```

cmd/api/main.go
```diff
@@ -187,28 +179,39 @@ func main() {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 	mux := runtime.NewServeMux()
-	err = v1alpha2pb.RegisterResultsHandlerFromEndpoint(ctx, mux, ":"+serverConfig.GRPC_PORT, opts)
+	err = v1alpha2pb.RegisterResultsHandlerFromEndpoint(ctx, mux, ":"+serverConfig.SERVER_PORT, opts)
```
adambkaplan, line 182: nit - I would make it clearer that `mux` is the http/REST handler by renaming it to `httpMux`

pkg/api/server/config/config.go
```diff
@@ -40,7 +39,7 @@ func Get() *Config {
 	viper.SetConfigName("config")
 	viper.SetConfigType("env")
 	viper.AddConfigPath("/etc/tekton/results")
-	viper.AddConfigPath("config/env")
+	viper.AddConfigPath("config/base/env")
```
adambkaplan, line 42: I think this could break deployments - is this meant to facilitate contributor testing?

adambkaplan (commented):
/approve

A few code nits (not blocking), and a question about the viper config path change.

/hold

I do have a concern that we don't have any e2e testing of the REST endpoint. The current tests appear to test via gRPC, so at least we can verify that works.

avinal:
Working fine in my tests: 
image

cmd/api/main.go
sayan-biswas: Agree, renamed.

pkg/api/server/config/config.go
sayan-biswas: Yes, that's right. Since the `config` directory was restructured into overlays in a previous PR, running the server directly from IDE was searching for the file in the `config/env` directory, which doesn't exist now.
This will NOT impact deployment, because deployment is using the absolute path `/etc/tekton/results` to mount config.
https://github.com/tektoncd/results/blob/2b03ffa8c753be7c07a67f9b6d78ac49c3370dba/config/base/api.yaml#L54

sayan-biswas:
@adambkaplan @enarha 
Some changes were done in general and to rebase, need review and lgtm again.

PR #419, opened 2023-03-29
Adding namespace setting

### LINKED ISSUE #417: namespace scoped tekton results
### Feature request

There should be a Watcher Parameter Setting to run the Watcher namespace scoped so all PipelineRuns, TaskRuns and Logs are watched and stored only namespace scoped.

### Use case

We plan to run in each namespace in it's own Tekton-Results Api and Watcher instance with it's own Schema. Each project has it's own build namespace, therefore it's own Buildlogs, and PipelineRuns and should also have it's own Tekton Results setup. That's why we want to separate all. So I know from Java Fabric8 Api that I can scope the Api to the namespace. That is actually all we need. Have a setting to scope the Watcher to the namespace only. The Rest we will do ourself. Role and Binding of the service account and so on.

Each Tekton Namespace has then it's segregation of Data. Namespace A cannot read PipelineRun Infos from Namespace B and only read it's own data.

manuelwallrapp:
# Changes

Added a namespace setting to define a specific namespace the Watcher should listen to. If the setting is not set, the Watcher will watch all namespaces.

Fixes #417 

# Submitter Checklist

# Release Notes

For pull requests with a release note:

```release-note
added namespace property to allow Watcher only watch a specific namespace
```

### DIFF
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -63,7 +63,7 @@ var (
 	logsAPI                 = flag.Bool("logs_api", true, "Disable sending logs. If not set, the logs will be sent only if server support API for it")
 	labelSelector           = flag.String("label_selector", "", "Selector (label query) to filter objects to be deleted. Matching objects must satisfy all labels requirements to be eligible for deletion")
 	requeueInterval         = flag.Duration("requeue_interval", 10*time.Minute, "How long the Watcher waits to reprocess keys on certain events (e.g. an object doesn't match the provided selectors)")
-	namespace               = flag.String("namespace", corev1.NamespaceAll, "Should the Watcher only watch a single namespace, then this values needs to be set to the namespace name otherwise leave it empty.")
+	namespace               = flag.String("namespace", corev1.NamespaceAll, "Should the Watcher only watch a single namespace, then this value needs to be set to the namespace name otherwise leave it empty.")
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


adambkaplan:
Ouch - this got bit by a race condition that was previously fixed, but was re-added.

Filed https://github.com/tektoncd/results/pull/424 to remove the failing unit test.

alan-ghelardi:
@manuelwallrapp  I left a minor suggestion for fixing a typo in the flag's usage, but I'm signaling that the patch looks good to me.
/lgtm

manuelwallrapp:
@alan-ghelardi thank you, but I can't see your suggestion?

cmd/watcher/main.go
```diff
@@ -63,6 +63,7 @@ var (
 	logsAPI                 = flag.Bool("logs_api", true, "Disable sending logs. If not set, the logs will be sent only if server support API for it")
 	labelSelector           = flag.String("label_selector", "", "Selector (label query) to filter objects to be deleted. Matching objects must satisfy all labels requirements to be eligible for deletion")
 	requeueInterval         = flag.Duration("requeue_interval", 10*time.Minute, "How long the Watcher waits to reprocess keys on certain events (e.g. an object doesn't match the provided selectors)")
+	namespace               = flag.String("namespace", corev1.NamespaceAll, "Should the Watcher only watch a single namespace, then this values needs to be set to the namespace name otherwise leave it empty.")
```
alan-ghelardi, line 66:
```suggestion
	namespace               = flag.String("namespace", corev1.NamespaceAll, "Should the Watcher only watch a single namespace, then this value needs to be set to the namespace name otherwise leave it empty.")
```

alan-ghelardi (commented):
Apologies @manuelwallrapp. I forgot to submit the suggestion.

adambkaplan:
/kind feature

I was initially reluctant to approve this (and associated feature request) because of concerns regarding long term maintainability. Most Tekton controllers operate by watching all namespaces, and single namespace mode isn't something that is supported by things like the Tekton Operator.

However, Tekton Pipelines (which is what we really care about) supports a "namespace only" mode that is not strongly documented beyond code/cli. And it's a feature that has existed almost forever - see https://github.com/tektoncd/pipeline/issues/2111.

adambkaplan (approved):
/lgtm

PR #495, opened 2023-06-07
Make cel2sql generic using View abstraction

RafaeLeal:
# Changes

This PR makes the cel2sql package more generic, introducing a new `View` abstraction. This allows us to decouple what we expose for the Results' API `filters` and the database structure.

I believe it also makes changes in our database safer since we don't have to think of all kinds of filters that would use each column of the db table. The idea is to make the changes and support in the CEL `filters` more deliberate.

# Submitter Checklist

# Release Notes


pkg/api/server/cel/view/records.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/tektoncd/results/pkg/api/server/cel2sql"
+)
+
+func NewRecordsView() (*cel2sql.View, error) {
+	view := &cel2sql.View{
+		Constants: map[string]cel2sql.Constant{},
+		Fields: map[string]cel2sql.Field{
+			"parent": {
+				CELType: cel.StringType,
+				SQL:     `parent`,
```
alan-ghelardi, line 14: Is this field the column name in the corresponding table? If so, I'd rename it to make the intent clearer.

pkg/api/server/cel/view/records_test.go
alan-ghelardi, line 1: I didn't understand why exactly these tests are exercising the conversions if, in practice, they aren't being performed by the records.go file. This seems counterintuitive for collaborators.

pkg/api/server/cel/view/results_test.go
alan-ghelardi, line 1: Same as the previous comment. This doesn't seem to be a good place to perform those tests, since the results.go file doesn't perform any conversion. I think that keeping the tests in the package responsible for performing the conversion is a better choice.

pkg/api/server/cel2sql/convert.go
```diff
@@ -20,9 +20,14 @@ import (
 	"github.com/google/cel-go/cel"
 )
 
-// Convert takes CEL expressions and attempt to convert them into Postgres SQL
-// filters.
-func Convert(env *cel.Env, filters string) (string, error) {
+// ConvertView takes a View and CEL expressions and attempt to convert them into
+// Postgres SQL filters.
+func ConvertView(view *View, filters string) (string, error) {
```
alan-ghelardi, line 25: I wouldn't change the name of the function. Ultimately, what is being converted is the CEL expression and not the view, which seems to be merely a container to represent the mapping among the database table and the object.

pkg/api/server/cel2sql/convert_test.go
```diff
@@ -18,265 +18,44 @@ import (
 	"errors"
 	"testing"
 
-	"github.com/tektoncd/results/pkg/api/server/cel"
-
+	"github.com/google/cel-go/cel"
 	"github.com/google/go-cmp/cmp"
 )
 
-func TestConvertRecordExpressions(t *testing.T) {
```
alan-ghelardi, line 26: As mentioned previously, I recommend keeping these tests here  instead of moving them to a different package.

pkg/api/server/cel2sql/doc.go
```diff
@@ -0,0 +1,3 @@
+// this package provides a generic interface to transform CEL expression into
```
alan-ghelardi, line 1: By convention, start the comment with "package cel2sql provides...".

pkg/api/server/cel2sql/select.go
```diff
@@ -35,13 +37,32 @@ func (i *interpreter) translateToJSONAccessors(fieldPath []fmt.Stringer) {
...
+	}
+	return s.String()
+}
+
+// translateIntoStruct
+func (i *interpreter) translateIntoStruct(fieldPath []fmt.Stringer) error {
+	namer := &schema.NamingStrategy{}
+	rawSql := getRawString(fieldPath[0])
+	rawField := getRawString(fieldPath[1])
+	sqlTemplate, err := template.New("").Parse(rawSql)
```
alan-ghelardi, line 55: Are there template placeholders being interpolated somewhere?

pkg/api/server/cel2sql/view_types.go
alan-ghelardi, line 1: Please, add doc comments to the main types and functions in this file to help collaborators to understand their purpose and how they can be extended/evolved.

alan-ghelardi:
@RafaeLeal  I left a few comments mostly about style and some parts of the code that I'd prefer to leave as before (tests, for example). You need to rebase the pr and write down release notes.

pkg/api/server/cel/view/records.go
RafaeLeal: Not necessarily will be a column.. it can be an operation like the concatenation of multiple columns or values

pkg/api/server/cel2sql/select.go
RafaeLeal: Yes, the summary is configured like this:

```
"summary": {
	...
	SQL:        `recordsummary_{{.Field}}`,
},
```

pkg/api/server/cel2sql/view_types.go
RafaeLeal: 👍

pkg/api/server/cel/view/records_test.go
RafaeLeal: I've initially moved to this file to avoid a cyclic dependency.. 
I've changed to different approach by having a test view that has all the fields that we need to test

pkg/api/server/cel2sql/convert.go
RafaeLeal: 👍

adambkaplan (approved):
/approve

The intent here is good - this lays a good foundation for making cel2sql an external library.

I recommend that the changes to `convert_test.go` be reverted, and replaced with only the necessary refactorings to convert the existing tests to use the new View type. The current set of changes make it difficult to verify that our test logic remains in place (and this PR doesn't break anything).

pkg/api/server/cel2sql/convert_test.go
```diff
@@ -18,234 +18,237 @@ import (
...
 
-	"github.com/tektoncd/results/pkg/api/server/cel"
-
+	"github.com/google/cel-go/cel"
 	"github.com/google/go-cmp/cmp"
+	resultspb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
 )
 
-func TestConvertRecordExpressions(t *testing.T) {
+func newTestView() *View {
```
alan-ghelardi, line 26: Couldn't we use the actual view rather than building a new one for testing? Asking because by doing so, we might forget  to configure the actual view properly and do that only in the test one or vice-versa and ultimately, we might end up not exercising the actual interpreter's logic.

Issue #522, opened 2023-07-11
Postgres STS fails to start - mkdir: cannot create directory ‘/bitnami/postgresql/data’


kbristow:
# Expected Behavior
Postgres created as part of the release manifests would start up successfully.

# Actual Behavior
The postgres pod does not become healthy. The pod fails and enters crashloopbackoff with the following error in the logs:

```
mkdir: cannot create directory ‘/bitnami/postgresql/data’: Permission denied
```

This appears similar to this issue: https://github.com/bitnami/charts/issues/1210

Investigating the recommendation [here](https://github.com/bitnami/charts/issues/1210#issuecomment-1179827873) i added an init container that looks as follows which resolves the issue:

```
initContainers:
  - name: init-chmod-data
    image: docker.io/bitnami/bitnami-shell:11-debian-11-r130
    imagePullPolicy: "IfNotPresent"
    resources:
      limits: {}
      requests: {}
    command:
      - /bin/sh
      - -ec
      - |
        chown 1001:1001 /bitnami/postgresql
        mkdir -p /bitnami/postgresql/data
        chmod 700 /bitnami/postgresql/data
        find /bitnami/postgresql -mindepth 1 -maxdepth 1 -not -name "conf" -not -name ".snapshot" -not -name "lost+found" | \
          xargs -r chown -R 1001:1001
    securityContext:
      runAsUser: 0
    volumeMounts:
      - name: postgredb
        mountPath: /bitnami/postgresql
```

I took the above from the output of running the below (with slight modifications):
```
helm template my-release oci://registry-1.docker.io/bitnamicharts/postgresql --set volumePermissions.enabled=true
```

Note that I am using EKS with ebs volumes and a storage class as below if useful:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
allowVolumeExpansion: true
provisioner: ebs.csi.aws.com
reclaimPolicy: Retain
volumeBindingMode: WaitForFirstConsumer
```

# Steps to Reproduce the Problem

1. Install Tekton Results (v0.7.0) on eks 1.23 with ebs volume
2. Observe error with postgres pod

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

```
v1.27.3
```

- Tekton Pipeline version:

```
v0.38.4
```

xinnjie:
Did you ever create PVC to interact with Postgres manually? This may cause different file permission since your reclaim policy of storage class is `Retain`. 

[...]

PR #528, opened 2023-07-15
Fix file permission mismatch for Postgres statefulset

xinnjie:
# Changes
This fix the issue https://github.com/tektoncd/results/issues/522
Kubelet will modify the ownership and permissions of volume to 1001 which Postgres uses after setting `securityContext.fsGroup`.
This pr will helper users who want to try out Results not encounter file permission issues. 

# Submitter Checklist
/kind bug

# Release Notes

### DIFF
--- a/config/components/local-db/201-sql-deployment.yaml
+++ b/config/components/local-db/201-sql-deployment.yaml
@@ -64,6 +64,8 @@ spec:
               - ALL
             add:
               - NET_BIND_SERVICE
+      securityContext:
+        fsGroup: 1001
   volumeClaimTemplates:
   - metadata:
       name: postgredb


config/components/local-db/201-sql-deployment.yaml
```diff
@@ -64,6 +64,8 @@ spec:
               - ALL
             add:
               - NET_BIND_SERVICE
+      securityContext:
+        fsGroup: 1001
```
sayan-biswas, line 68: I feel this should go in document rather than in default config because this will not work in platform with default restrictions like Openshift.
That can break the tekton operator deployment of results. Correct me if I am wrong @khrm .

xinnjie: What default restrictions Openshift have about `securityContext. fsGroup`?

sayan-biswas: I deployed this in Openshift earlier today with default SCC and it didn't bring the pod up. I can check again later.

sayan-biswas: @khrm Does this have any impact on the tekton operator deployment of tekton results in Openshift clusters?

90dy:
Hi,
I continue to have this issue, I think this PR needs to be merged

PR #588, opened 2023-08-31
add instructions for handling large log files

avinal:
# Changes

- log size is limited by the `LOGS_BUFFER_SIZE` and it is not logical to
  increase it too high
- Reaching this limit causes the response to be broken in chunks where
  each chunk is a valid JSON/proto response but the complete response is
  no longer a valid JSON/proto
- This must be handled from the client size by parsing each chunk and
  then combining the data field to retrieve the complete log
- Added example for handling such case using bash commands

Signed-off-by: Avinal Kumar <avinal@redhat.com>

# Submitter Checklist

# Release Notes

```release-note
none
```

### DIFF
--- a/docs/large-log-files.md
+++ b/docs/large-log-files.md
@@ -1,0 +1,118 @@
+# Fetching large log files using Tekton Results API
+
+Run logs can get huge, Tekton Results help archive logs away from the cluster
+and free up space. You can fetch these logs using Tekton Results API. There are
+multiple ways to fetch logs from Tekton Results API. You can use gRPC or REST API.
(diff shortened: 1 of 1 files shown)


avinal:
cc: @khrm @enarha @sayan-biswas

khrm (commented):
/approve

adambkaplan:
/lgtm

This suggests that we need a better mechanism for serving large logs as a stream of text. Requiring all of the data to be streamed before base64 decoding will not scale to mutli-gigabyte log files.

PR #625, opened 2023-09-28
Make sure gRPC connection is closed

### LINKED ISSUE #624: GRPC Stream Leak on Error
Within RHTAP there have been multiple occasions where the tekton results watcher has hung and stopped processing results. In each case the last message in the log was a message similar to:

`
{"level":"error","ts":"2023-09-27T11:48:52.977Z","logger":"watcher","caller":"dynamic/dynamic.go:326","msg":"Error streaming log","knative.dev/traceid":"53b3d4b0-823c-4aec-a6e7-5d51f4ff8d77","knative.dev/key":"build-templates-e2e/verify-enterprise-contract-run-bgn9f","results.tekton.dev/kind":"PipelineRun","namespace":"build-templates-e2e","kind":"PipelineRun","name":"verify-enterprise-contract-run-bgn9f","error":"error reading from tkn reader: pipelineruns.tekton.dev \"verify-enterprise-contract-run-bgn9f\" not found","stacktrace":"github.com/tektoncd/results/pkg/watcher/reconciler/dynamic.(*Reconciler).sendLog.func1\n\t/opt/app-root/src/pkg/watcher/reconciler/dynamic/dynamic.go:326"}
`

And then nothing. 

I had a look at the relevant code and it appears that in this case a GRPC stream is opened to stream the logs, however in the error return case it is never closed. My hypothesis is that after this happens a number of times we exhaust the number of available GRPC connections/streams and the watcher hangs.

stuartwdouglas:
# Changes

In the error case the gRPC stream is closed. By using a new context and cancelling it when the function is finished we should guarentee that the gRPC context is closed.

Fixes #624

I have not tested this locally as I don't have the time to setup a local results build, and I am not sure how to test it with unit tests.

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -343,6 +343,8 @@ func (r *Reconciler) sendLog(ctx context.Context, o results.Object) error {
 
 func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, logName string) error {
 	logger := logging.FromContext(ctx)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	logsClient, err := r.resultsClient.UpdateLog(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to create UpdateLog client: %v", err)


khrm (commented):
/ok-to-text

khrm:
We probably need an e2e test to check this. Sure I would look into this.
/assign @khrm

stuartwdouglas:
Does this need anything else from my end?

enarha:
@stuartwdouglas No. And thank you for the contribution! Let me try test that.

adambkaplan:
/lgtm

It is pretty hard to add unit or e2e tests around this sort of thing - we're really only seeing this behavior with real operations at scale.

PR #626, opened 2023-10-03
improve watcher logs

ramessesii2:
- improve watcher logs messages

/kind cleanup

# Changes

# Submitter Checklist

# Release Notes

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -108,7 +108,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, o results.Object) error {
 	timeTakenField := zap.Int64("results.tekton.dev/time-taken-ms", time.Since(startTime).Milliseconds())
 
 	if err != nil {
-		logger.Debugw("Error upserting record", zap.Error(err), timeTakenField)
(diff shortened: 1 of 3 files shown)


ramessesii2:
cc @sayan-biswas

enarha:
@ramessesii2 personally I like we have the name and the namespace as separate fields as having the combo in "knative.dev/key" is less obvious. I agree the kind is probably redundant as it's quite obvious from the "results.tekton.dev/kind" field. That's IMO. Having said that, I do not have strong objection, so if others prefer the shorter records, then OK with me.

avinal (changes requested):
+1 to Emil. Although it may seem that the fields are redundant. Having explicit common fields like name, namespace, and kind is preferable and direct. In an ideal case, a user/log parser will look out for well-known fields than a less common field injected by the knative framework.

So, IMO we should leave them as it is. The message improvement is necessary and appreciated, and if you find more instances for improvement, please do that.

ramessesii2:
@enarha @avinal given your reviews, I propose having the keys 'namespace' & 'name' uniform throughout the watcher logs. Right now, it's sporadic. 
Also, we seem to agree having "results.tekton.dev/kind" is enough for marking a `kind` in logs, so I'll get rid of `kind` keys.

avinal (approved):
/approve
/lgtm

PR #642, opened 2023-10-18
Implementation of Summary/Aggregation API in Tekton results

avinal:
## Changes

- add Summary proto definition
- implementation for summary of list of records
- docs for summary/aggregation API
- added clusterrole permissions
- Refer: tektoncd/community#1086

```release-note
Added a new api endpoint for fetching summary and aggregation for list of records. Now you can directly get aggreagation data for a given list of records without fetching and processing on the client side.
```


pkg/api/server/v1alpha2/lister/aggregator.go
```diff
@@ -0,0 +1,200 @@
...
+func GetCount(query, countName string) aggregateFunc {
+	return func(db *gorm.DB) *gorm.DB {
+		return db.Select(db.Statement.Selects, "COUNT("+query+") AS "+countName)
+	}
+}
+
+type aggregateFunc func(db *gorm.DB) *gorm.DB
+
+var summaryFuncs = map[string]aggregateFunc{
+	"total":          GetCount("*", "total"),
```
khrm, line 160: Can we just make a single call? Like I did? Why don't we use TYPE as a group also?

khrm: Then we can get all these in the single call.

avinal: I actually took the inspiration from your code and then came up with this 😅. Practically, it is a single call only. You can enable DB logs and see them. 

This design lets us define and add new queries without adding a lot of code, but still maintains the code readability. Furthermore, we can use the same function for both records and results and even a combination of many functions.

pkg/api/server/v1alpha2/summary.go
```diff
@@ -0,0 +1,40 @@
...
+	if req.GetParent() == "" {
+		return nil, status.Error(codes.InvalidArgument, "parent missing")
+	}
+
+	parent, resultName, err := result.ParseName(req.GetParent())
+	if err != nil {
+		return nil, err
+	}
+
+	if err := s.auth.Check(ctx, parent, auth.ResourceRecords, auth.PermissionGet); err != nil {
```
sayan-biswas, line 25: Shouldn't this API use a different `Resource` for permission, otherwise it will impossible to give an user access to fetch Records and prevent the user from fetching the Summary at the same time.

pkg/api/server/v1alpha2/summary.go
```diff
@@ -0,0 +1,40 @@
...
+		return nil, err
+	}
+
+	recordAggregator, err := lister.OfRecordList(s.recordsEnv, parent, resultName, req)
+	if err != nil {
+		return nil, err
+	}
+
+	agg, err := recordAggregator.Aggregate(ctx, s.db)
```
sayan-biswas, line 34: I don't see the `context` being used in any of the child methods under this call tree. So we cal directly create a db with context and pass that only, instead of passing the context separately.

pkg/api/server/v1alpha2/lister/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+}
+
+func getCount(query, countName string) aggregateFunc {
+	return func(db *gorm.DB) *gorm.DB {
+		return db.Select(db.Statement.Selects, "COUNT("+query+") AS "+countName)
+	}
+}
+
+type aggregateFunc func(db *gorm.DB) *gorm.DB
```
sayan-biswas, line 164: Declaring the type before using helps in code readability IMO.

pkg/api/server/v1alpha2/lister/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+		fn, ok := summaryFuncs[q]
+		if !ok {
+			return nil, status.Errorf(codes.InvalidArgument, "invalid aggregate query: %s", q)
+		}
+		fns = append(fns, fn)
+	}
+	return fns, nil
+}
+
+func (a *Aggregator) applyAggregateFunc(ctx context.Context, db *gorm.DB) *gorm.DB {
```
sayan-biswas, line 191: In a lot of these methods we are already sending the `gorm.DB` struct as pointer and return the same pointer.

pkg/api/server/v1alpha2/lister/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+	return func(db *gorm.DB) *gorm.DB {
+		groupSelect := "EXTRACT(EPOCH FROM DATE_TRUNC('" + group + "', " + query + ")) AS group_value"
+		return db.Select(db.Statement.Selects, groupSelect).Group("group_value")
+	}
+}
+
+func getStatus(query, reason string) aggregateFunc {
+	return func(db *gorm.DB) *gorm.DB {
+		statusSelect := ""
+		if reason == "Unknown" {
```
sayan-biswas, line 149: Is there an established definition of `Unknown`? As per tekton pipelines API, there is another status `TaskrunTimeout` and `PipelinerunTimeout`. Moreover PipelineRun API have and extra status `Completed`, when the pipelinerun completes skipped tasks.

pkg/api/server/v1alpha2/lister/aggregator.go
```diff
@@ -0,0 +1,208 @@
...
+
+// buildQuery applies filters
+func (a *Aggregator) buildQuery(ctx context.Context, db *gorm.DB) (*gorm.DB, error) {
+	var err error
+	db = db.WithContext(ctx)
+
+	for _, builder := range a.queryBuilders {
+		db, err = builder.build(db)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
```
sayan-biswas, line 43: This statement will expose internal error description in the response.

sayan-biswas (changes requested):
Generally looks good, minor code changes required along with godoc annotations and unit tests.

pkg/api/server/v1alpha2/lister/aggregator.go
avinal: We are only choosing certain statues. So we can omit other fields.

pkg/api/server/v1alpha2/lister/aggregator.go
avinal: yes, I do see an improvement scope here by including this as a part of the struct. Will do that.

pkg/api/server/v1alpha2/lister/aggregator.go
avinal: Sure I will move it up

pkg/api/server/v1alpha2/lister/aggregator.go
avinal: this build statement puts error related to compilation of filters, so we can either put a generic statement here or put the compilation error.

vdemeester:
@avinal needs a release-note entry I think 👼🏼 
/approve

proto/v1alpha2/api.proto
```diff
@@ -163,6 +169,18 @@ message GetResultRequest {
     }];
 }
 
+message RecordListSummaryRequest {
+  string parent = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (google.api.resource_reference) = {
+      child_type: "tekton.results.v1alpha2/Record"
```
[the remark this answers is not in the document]
avinal: what would be a good type then?

sayan-biswas (approved):

khrm (commented):
/lgtm

Issue #514, opened 2023-06-26
Individual log is not retrievable via the name returned in the log list


iainsproat:
# Expected Behavior

```
curl -k https://speckle.internal:8088/apis/results.tekton.dev/v1alpha2/parents/bfeb4b5d45/results/47373980-cbcb-4b57-8746-40cc5fd75649/logs
```

This results in the following as expected:
```
{"records":[{"name":"bfeb4b5d45/results/47373980-cbcb-4b57-8746-40cc5fd75649/logs/77a71268-1422-39f4-8b37-8e5264c6f2ae", "id":"58f21208-d056-4506-a1c8-124b3e34c830", "uid":"58f21208-d056-4506-a1c8-124b3e34c830", "data":{"type":"results.tekton.dev/v1alpha2.Log", "value":"base64encodedlogs"}, "etag":"58f21208-d056-4506-a1c8-124b3e34c830-1687799787144731088", "createdTime":"2023-06-26T17:16:27.144731Z", "createTime":"2023-06-26T17:16:27.144731Z", "updatedTime":"2023-06-26T17:16:27.144731Z", "updateTime":"2023-06-26T17:16:27.144731Z"}], "nextPageToken":""}`
```

I expect that using the `.records[0].name` would return the individual log record that can be seen in the list, e.g.:

```
curl -k https://speckle.internal:8088/apis/results.tekton.dev/v1alpha2/parents/bfeb4b5d45/results/47373980-cbcb-4b57-8746-40cc5fd75649/logs/77a71268-1422-39f4-8b37-8e5264c6f2ae
```

I may be doing something incorrect here, but it seems to match the API documentation.

What I expect to have returned is the following:

```
{"name":"bfeb4b5d45/results/47373980-cbcb-4b57-8746-40cc5fd75649/logs/77a71268-1422-39f4-8b37-8e5264c6f2ae", "id":"58f21208-d056-4506-a1c8-124b3e34c830", "uid":"58f21208-d056-4506-a1c8-124b3e34c830", "data":{"type":"results.tekton.dev/v1alpha2.Log", "value":"base64encodedlogs"}
```

# Actual Behavior

Retrieving the `.records[0].name` and attempting to access the resource results in the log record not being found.

```
[...]

PR #672, opened 2023-11-22
Add owner references check optional in watcher

sayan-biswas:
As per discussion in #648 

# Changes
This PR makes the owner reference check in watcher optional.

# Submitter Checklist

# Release Notes

```release-note
"check_owner" command line flag allows the Owner References check to be optional while deleting resources.
```

### DIFF
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -66,6 +66,7 @@ var (
 	labelSelector           = flag.String("label_selector", "", "Selector (label query) to filter objects to be deleted. Matching objects must satisfy all labels requirements to be eligible for deletion")
 	requeueInterval         = flag.Duration("requeue_interval", 10*time.Minute, "How long the Watcher waits to reprocess keys on certain events (e.g. an object doesn't match the provided selectors)")
 	namespace               = flag.String("namespace", corev1.NamespaceAll, "Should the Watcher only watch a single namespace, then this value needs to be set to the namespace name otherwise leave it empty.")
+	checkOwner              = flag.Bool("check_owner", true, "If enabled, owner references will be checked while deleting objects")
 )
 
 func main() {
@@ -97,6 +98,7 @@ func main() {
 		DisableAnnotationUpdate:      *disableCRDUpdate,
 		CompletedResourceGracePeriod: *completedRunGracePeriod,
 		RequeueInterval:              *requeueInterval,
+		CheckOwner:                   *checkOwner,
 	}
 
 	if selector := *labelSelector; selector != "" {
--- a/docs/watcher/README.md
+++ b/docs/watcher/README.md
@@ -57,3 +57,9 @@ metadata:
     results.tekton.dev/recordSummaryAnnotations: |-
       {"foo": "bar"}
 ```
+
+## Resource Deletion
+
+When the command line flag is `completed_run_grace_period` is set to any value other than `0`, resources will be deleted after the specified duration in the flag, calculated from the time of completion. If the value is < `0`, Runs will be deleted immediately after completion or failure.
+
+The flag `check_owner` allows additional check before deleting a resource. If set `true`, resources with any owner references set will not be deleted. When the flag is `false`, owner references will be not be checked before deletion. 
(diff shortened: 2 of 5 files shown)


cmd/watcher/main.go
```diff
@@ -66,6 +66,7 @@ var (
 	labelSelector           = flag.String("label_selector", "", "Selector (label query) to filter objects to be deleted. Matching objects must satisfy all labels requirements to be eligible for deletion")
 	requeueInterval         = flag.Duration("requeue_interval", 10*time.Minute, "How long the Watcher waits to reprocess keys on certain events (e.g. an object doesn't match the provided selectors)")
 	namespace               = flag.String("namespace", corev1.NamespaceAll, "Should the Watcher only watch a single namespace, then this value needs to be set to the namespace name otherwise leave it empty.")
+	checkOwner              = flag.Bool("check_owner", true, "If enabled, owner references will be checked while deleting objects")
```
Roming22, line 69: I thought we said that the default behavior would match the tekton pruner behavior (in which case the default should be `false`). Am I mistaken?

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -182,7 +182,7 @@ func (r *Reconciler) deleteUponCompletion(ctx context.Context, o results.Object)
 		return nil
 	}
 
-	if ownerReferences := o.GetOwnerReferences(); len(ownerReferences) > 0 {
+	if ownerReferences := o.GetOwnerReferences(); r.cfg.CheckOwner && len(ownerReferences) > 0 {
```
Roming22, line 185: I thought that one of the reason the check was put in place was to avoid deleting taskruns before the pipelineruns that owns them had the time to process the taskrun's logs.
If that's correct, additional logic needs to be added to cancel the deletion in the case one of the ownerRef is of type PipelineRun, even if checkOwner is set to false.

cmd/watcher/main.go
enarha: I believe @sayan-biswas tried to preserve the existing behavior as per https://github.com/tektoncd/results/discussions/648#discussion-5776305, but I agree it makes more sense to change that to false.

docs/watcher/README.md
```diff
@@ -57,3 +57,9 @@ metadata:
     results.tekton.dev/recordSummaryAnnotations: |-
       {"foo": "bar"}
 ```
+
+## Resource Deletion
+
+When the command line flag is `completed_run_grace_period` is set to any value other than `0`, resources will be deleted on completion. If < `0`, Runs will be deleted immediately.
```
enarha, line 63: That sounds a bit confusing to me. If the value is set to 100, wouldn't the runs be deleted after that grace period, instead of "will be deleted on completion"?

sayan-biswas: Yes, I'll change the description here.

cmd/watcher/main.go
sayan-biswas: If we make it a default behaviour, then it would be a BREAKING change for the existing tekton results users. Kubernetes already implements the cascading deletion feature through owner references by default. IMO we should  keep that behaviour unless some users wants to override that, and this flag will help in that.

pkg/watcher/reconciler/dynamic/dynamic.go
sayan-biswas: This check was implemented in #255 to improve stability in deletion, and it was not related to the processing of taskrun logs. 
For example, consider the situation if only a taskrun is executed (ie. without a parent pipelinerun), in that case there won't be any owner references and hence the controller won't wait for the deletion before log streaming is completed or failed.

Having said that, I agree that there should be additional logic to handle the log streaming situation, but it's not a scope of this PR, as that logic is not related to `ownerRef`. There is a open bug for this, #514.

cmd/watcher/main.go
enarha: Keeping backwards compatibility is important as well. After considering that again, I tend to agree with preserving the existing behavior. Users who want to change that, could use the new flag to do so.

docs/watcher/README.md
sayan-biswas: Changed.

cmd/watcher/main.go
Roming22: OK

PR #690, opened 2024-01-05
Fix deletion when TaskRun is part of PipelineRun

### LINKED ISSUE #689: TaskRuns gets deleted  before PipelineRun completes if check_owner flag is disabled
# Expected Behavior

TaskRuns which are part of a PipelineRun should not get deleted separately. Instead should wait for parent deletion.

# Actual Behavior

TaskRuns gets deleted  before PipelineRun completes if check_owner flag is disabled

Note:
Right now this is a bug because the Pipeline Controller doesn't handle the edge case where the child TaskRun can be deleted without deleting the parent PipelineRun, and as a result the deletion causes the PipelineRun object to get updated with a "Pending" status for that task.

sayan-biswas:
Fixes #689 

# Changes

Prevent deletion of any object OwnerRef check is disabled and yet the object has a PipelineRun as owner ref.

This is needed because the Pipeline Controller doesn't handle the edge case where the child TaskRun can be deleted without deleting the parent PipelineRun, and as a result the deletion causes the PipelineRun object to get updated with a "Pending" status for that task.

If an aggressive pruning is required, this can be removed in future once the above scenario is handled by the Pipeline Controller.

# Submitter Checklist

# Release Notes

```release-note
This change will prevent the controller from directly pruning TaskRuns which are initiated by a PipelineRun. Instead these TaskRuns will be deleted only when the owner PipelineRun is deleted.
```

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic_test.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic_test.go
@@ -335,8 +335,11 @@ func TestReconcile_TaskRun(t *testing.T) {
 		taskrun.OwnerReferences = []metav1.OwnerReference{{
 			APIVersion: "v1",
 			Kind:       "test",
-			Name:       "test-parent",
+			Name:       "test-owner",
 		}}
+
+		// delete taskrun before creating
+		trclient.Delete(ctx, taskrun.GetName(), metav1.DeleteOptions{})
 
 		// Recreate the object to retest the deletion
 		if _, err := trclient.Create(ctx, taskrun, metav1.CreateOptions{}); err != nil {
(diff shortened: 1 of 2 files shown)


Roming22 (approved):

vdemeester (approved):

gabemontero:
> This is needed because the Pipeline Controller doesn't handle the edge case where the child TaskRun can be deleted without deleting the parent PipelineRun, and as a result the deletion causes the PipelineRun object to get updated with a "Pending" status for that task.

to be clear @sayan-biswas , and apologies if you feel it is too obvious, but rather then updating with a "pending" status for the task, can you explicitly state what updates if any the pipeline controller should make on the PipelineRun object for the task? some status of "deleted" ?

Among other things I am curious because of the need to handle the deletion of taskruns owned by pipelineruns that are still alive  in some other scenarios.

Also, coming in new to this, I am a bit curious why we might not ever want to always honor the owner ref ... is it to simply speed up pruning ?

sayan-biswas:
@gabemontero As per my understanding, the pipeline controller propagates the status of the child TaskRuns to the parent PipelineRun object when it changes. But in this case, by design, deletion of TaskRun will trigger a reconciliation but I don't get what is the point of updating the parent PipelineRun object to "Pending", is it not possible to keep it as it is. 
Its been a long time I've visited the tekton pipeline project code, let me find out what is causing this, then I'll have more understanding.

Roming22:
> Also, coming in new to this, I am a bit curious why we might not ever want to always honor the owner ref ... is it to simply speed up pruning ?

We have PipelineRuns that have an ownerRef. Those PipelineRuns are not being pruned, and end up polluting the namespace.

pkg/watcher/reconciler/dynamic/dynamic_test.go
```diff
@@ -353,16 +356,47 @@ func TestReconcile_TaskRun(t *testing.T) {
...
+		// Recreate the object to retest the deletion
+		if _, err := trclient.Create(ctx, taskrun, metav1.CreateOptions{}); err != nil {
+			t.Fatal(err)
+		}
+
+		if err := r.Reconcile(ctx, taskrun); err != nil {
+			t.Fatal(err)
+		}
+
+		// Make sure that the resource no longer exists
```
adambkaplan, line 381: Fix comment - resource should exist.
There are other identical comments that should be corrected - appears to be a copy/paste issue.

```suggestion
		// Make sure the TaskRun is not deleted
```

adambkaplan (changes requested):
Code changes look good - I found that some of the comments in the test code need to be fixed as the comment does not match the logic in the test.

Otherwise my concerns fall into the general "code hygiene" category:

1. Include a body in the commit message that explains the _what and why_ of the change. See the [Tekton commit message standards](https://github.com/tektoncd/community/blob/main/standards.md#commit-messages) 
2. Fill in the release note of the pull request description - this is a behavior change that warrants a release note.

adambkaplan:
I'm setting aside user-facing docs concerns, as we really don't have great documentation today.

adambkaplan:
Does this fix #689?

pkg/watcher/reconciler/dynamic/dynamic_test.go
sayan-biswas: Done

sayan-biswas:
> Does this fix #689?

Yes.

adambkaplan (approved):
/lgtm

PR #703, opened 2024-02-05
[WIP] finalizer approach to fix race condition due to pruning in results watcher

ramessesii2:
# Changes

/kind bug

# Submitter Checklist

# Release Notes
```release-note
NONE
```

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -388,18 +388,38 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, 
 		return fmt.Errorf("error reading from tkn reader: %w", err)
 	}
 
-	errChanRepeater := make(chan error)
+	errChanRepeater := make(chan error, 100) //some stuff on the internet says buffered channels are better for GC
 	go func(echan <-chan error, o metav1.Object) {
-		writeErr := <-echan
-		errChanRepeater <- writeErr
-
-		_, err := writer.Flush()
-		if err != nil {
-			logger.Error(err)
+		select {
+		case <-ctx.Done():
+			logger.Warnw("Context done streaming log",
+				zap.String("namespace", o.GetNamespace()),
+				zap.String("name", o.GetName()),
+			)
+		case <-time.After(10 * time.Minute): //TODO could make this time configurable, but let's see how useful it is first
+			logger.Warnw("10 minute timer expired streaming log",
+				zap.String("namespace", o.GetNamespace()),
+				zap.String("name", o.GetName()),
+			)
+		case writeErr := <-echan:
+			errChanRepeater <- writeErr
+			var gofuncerr error
+			_, gofuncerr = writer.Flush()
+			if gofuncerr != nil {
+				logger.Error(gofuncerr)
+			}
+			if gofuncerr = logsClient.CloseSend(); gofuncerr != nil {
+				logger.Error(gofuncerr)
+			}
+			logger.Debugw("Flush in streamLogs done",
+				zap.String("namespace", o.GetNamespace()),
+				zap.String("name", o.GetName()),
+			)
 		}
-		if err = logsClient.CloseSend(); err != nil {
-			logger.Error(err)
-		}
+		logger.Debugw("Gofunc in streamLogs done",
+			zap.String("namespace", o.GetNamespace()),
+			zap.String("name", o.GetName()),
+		)
 	}(errChan, o)
 
 	// errChanRepeater receives stderr from the TaskRun containers.
@@ -409,6 +429,11 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, 
 		Out: writer,
 		Err: writer,
 	}, logChan, errChanRepeater)
+
+	logger.Debugw("Exiting streamLogs",
+		zap.String("namespace", o.GetNamespace()),
+		zap.String("name", o.GetName()),
+	)
 
 	return nil
 }
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -401,6 +401,7 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, 
 
 	// logctx is derived from ctx. Therefore, if ctx is cancelled (either explicitly through a call to its cancel
 	// function or when it reaches its deadline), logctx will be cancelled automatically.
+	// TODO: Implement configurable timeout based on user feedback analysis.
 	logctx, _ := context.WithTimeout(ctx, 10*time.Minute)
 
 	go func(ctx context.Context, echan <-chan error, o metav1.Object) {
@@ -413,24 +414,23 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, 
 			)
 		case writeErr := <-echan:
 			errChanRepeater <- writeErr
-			var gofuncerr error
-			_, gofuncerr = writer.Flush()
-			if gofuncerr != nil {
-				logger.Error(gofuncerr)
-			}
-			if gofuncerr = logsClient.CloseSend(); gofuncerr != nil {
-				logger.Error(gofuncerr)
-			}
-			logger.Debugw("Flush in streamLogs done",
-				zap.String("namespace", o.GetNamespace()),
-				zap.String("name", o.GetName()),
-			)
 		}
-		logger.Debugw("Gofunc in streamLogs done",
+		var gofuncerr error
+		_, gofuncerr = writer.Flush()
+		if gofuncerr != nil {
+			logger.Error(gofuncerr)
+		}
+		logger.Info("Flush in streamLogs done",
 			zap.String("namespace", o.GetNamespace()),
 			zap.String("name", o.GetName()),
 		)
-
+		if err = logsClient.CloseSend(); err != nil {
+			logger.Error(err)
+		}
+		logger.Info("Gofunc in streamLogs done",
+			zap.String("namespace", o.GetNamespace()),
+			zap.String("name", o.GetName()),
+		)
 		err = r.RemoveFinalizer(ctx, o)
 		if err != nil {
 			logger.Errorw("Error removing finalizer",
@@ -441,7 +441,7 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, 
 			return
 		}
 
-		logger.Debugw("Finalizer removed successfully",
+		logger.Info("Finalizer removed successfully",
 			zap.String("namespace", o.GetNamespace()),
 			zap.String("name", o.GetName()),
 		)
@@ -453,8 +453,8 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, 
 				zap.String("name", o.GetName()),
 				zap.Error(err),
 			)
-
 		}
+
 	}(logctx, errChan, o)
 
 	// errChanRepeater receives stderr from the TaskRun containers.
@@ -465,7 +465,7 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, 
 		Err: writer,
 	}, logChan, errChanRepeater)
 
-	logger.Debugw("Exiting streaming logs",
+	logger.Info("Exiting streaming logs",
 		zap.String("namespace", o.GetNamespace()),
 		zap.String("name", o.GetName()),
 	)
@@ -473,6 +473,14 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, 
 }
 
 func (r *Reconciler) AddFinalizer(ctx context.Context, o metav1.Object) error {
+	if ownerReferences := o.GetOwnerReferences(); len(ownerReferences) > 0 {
+		// do not add finalizer if the object is owned by a PipelineRun object
+		for _, or := range ownerReferences {
+			if or.Kind == "PipelineRun" {
+				return nil
+			}
+		}
+	}
 	finalizers := o.GetFinalizers()
 	for _, f := range finalizers {
 		if f == LogFinalizer {
@@ -481,7 +489,7 @@ func (r *Reconciler) AddFinalizer(ctx context.Context, o metav1.Object) error {
 	}
 	finalizers = append(finalizers, LogFinalizer)
 
-	patch, err := finalizerPatch(o, finalizers)
+	patch, err := finalizerPatch(finalizers)
 	if err != nil {
 		return fmt.Errorf("error adding results log finalizer: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -342,11 +356,6 @@ func (r *Reconciler) sendLog(ctx context.Context, o results.Object) error {
 					zap.Error(err),
 				)
 			}
-			logger.Debugw("Streaming log completed",
```
ramessesii2, line 345: This asynchronous functions never reaches this point if there is a successful streaming of logs while we do see this when there an error while streaming logs

khrm (commented):
/ok-to-test

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -81,8 +84,10 @@ func NewDynamicReconciler(rc pb.ResultsClient, lc pb.LogsClient, oc ObjectClient
 		resultsClient: results.NewClient(rc, lc),
 		objectClient:  oc,
 		cfg:           cfg,
-		// Always true predicate.
 		IsReadyForDeletionFunc: func(ctx context.Context, object results.Object) (bool, error) {
+			if LogsFinalizerExist(object, LogFinalizer) {
```
Roming22, line 88: I would think that this is not needed, as it is redundant with the `finalizer` pattern. 

I would expect the resource to be marked as being deleted, but the deletion to be defered by k8s until the last finalizer has been removed.

I could definitely be wrong though :wink:

ramessesii2: You're right deletion is deferred if the Finalizer is there.
So, prior to that change we could see the [following message](https://github.com/tektoncd/results/pull/703/files#diff-a48f98781edc438b09b76d3a97402d3730a04e55fffb4d049c08545841a4f7afR256) while debugging even if the resource is not deleted properly(due to Finalizer)  which is misleading.

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -388,19 +397,65 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType,
...
+				zap.String("name", o.GetName()),
+			)
+		case writeErr := <-echan:
+			errChanRepeater <- writeErr
+			var gofuncerr error
+			_, gofuncerr = writer.Flush()
+			if gofuncerr != nil {
+				logger.Error(gofuncerr)
+			}
+			if gofuncerr = logsClient.CloseSend(); gofuncerr != nil {
```
gabemontero, line 420: per comment from @adambkaplan in our internal discussion @ramessesii2 we want to execute the `CloseSend()` here regardless of which case in the select fires, meaning it needs to be moved out of this case, out of the select in fact, so that it is also executed when ctx.Done() is processed as well

gabemontero (changes requested):
The child context with timeout bit idea from @adambkaplan looks like how I expected it .... just the moving of the CloseSend is needed @ramessesii2

pkg/watcher/reconciler/dynamic/dynamic.go
gabemontero: i.e release the streaming resources regardless if the tkn client library is not properly reconciling the error channel used to block the goroutine from completing 

that at least would help slow things .... cleaning up the threads (or moving to a different model) could become a separate concern based on what you find as you drill down on your suspicion

gabemontero: So we've confirmed in an internal slack thread including @ramessesii2 @sayan-biswas and myself that the `streamLogs` code in the "main" thread needs to close the channel the goroutine is blocked on.

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -397,13 +397,14 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType,
...
 
-	errChanRepeater := make(chan error, 100) //some stuff on the internet says buffered channels are better for GC
+	errChanRepeater := make(chan error, 100)
 
 	// logctx is derived from ctx. Therefore, if ctx is cancelled (either explicitly through a call to its cancel
 	// function or when it reaches its deadline), logctx will be cancelled automatically.
 	logctx, _ := context.WithTimeout(ctx, 10*time.Minute)
 
 	go func(ctx context.Context, echan <-chan error, o metav1.Object) {
+		defer close(errChanRepeater)
```
gabemontero, line 407: good use of defer @ramessesii2 @sayan-biswas 

where are we closing the original `logChan` and `errChan` from the `reader.Read()`  ?

ramessesii2: logChan and errChan channels are receiver ends. When the channels ae no longer used, they will be garbage collected.
It's new subject for me, this is my reference - https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open

gabemontero: yeah there was a bit of back and forth on that thread @ramessesii2 but yes I think that is the end interpretation.

good find / research @ramessesii2 

that said, given the exchanges I just saw with @pmacik today in slack, I should probably make a more thorough pass of the changes here  .... admittedly I only focused on the mem leak bit initially.  

will do that today

thanks

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -388,19 +397,65 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType,
...
-	errChanRepeater := make(chan error)
-	go func(echan <-chan error, o metav1.Object) {
-		writeErr := <-echan
-		errChanRepeater <- writeErr
+	errChanRepeater := make(chan error, 100)
 
-		_, err := writer.Flush()
+	// logctx is derived from ctx. Therefore, if ctx is cancelled (either explicitly through a call to its cancel
+	// function or when it reaches its deadline), logctx will be cancelled automatically.
+	logctx, _ := context.WithTimeout(ctx, 10*time.Minute)
```
gabemontero, line 404: put a TODO comment about we are deferring on making this configurable for now until feedback from users uncovers the need for it

pkg/watcher/reconciler/dynamic/dynamic.go
gabemontero: OK @ramessesii2 your call to CloseSend minimally is still only in the `case writeErr := <-echan:` block

we need to call this as well if the `case <-ctx.Done():` block fire first ... i.e. there was some weird hang on the write, so we can abort and free up the resources

my initial assumption is the flush is irrelevant in this error case and case stay where it is, but do let us know if your testing uncovers a different conclusion

please update your PR so we can close this thread and have @pmacik 's next round of testing include this change

pkg/watcher/reconciler/dynamic/dynamic.go
gabemontero: ok @ramessesii2 @sayan-biswas I think I looked at this too quickly yesterday

the close of the `errChanRepeater` should not be in the goroutine that blocks on it !!

move the `defer close(errChanRepeater)` to immediately after the declaration of `errChanRepeater`  in your current line 400

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -388,19 +397,65 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType,
...
+			errChanRepeater <- writeErr
+			var gofuncerr error
+			_, gofuncerr = writer.Flush()
+			if gofuncerr != nil {
+				logger.Error(gofuncerr)
+			}
+			if gofuncerr = logsClient.CloseSend(); gofuncerr != nil {
+				logger.Error(gofuncerr)
+			}
+			logger.Debugw("Flush in streamLogs done",
```
gabemontero, line 424: while @pmacik is testing for us, let's temporarily change this to an Info level log

that way, if need be, we can look for these message in the results watcher log gathered in the artifacts from his e2e tests

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -388,19 +397,65 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType,
...
+				zap.String("namespace", o.GetNamespace()),
+				zap.String("name", o.GetName()),
+				zap.Error(err),
+			)
+			return
 		}
-		if err = logsClient.CloseSend(); err != nil {
-			logger.Error(err)
+
+		logger.Debugw("Finalizer removed successfully",
```
gabemontero, line 444: while @pmacik is testing for us, let's temporarily change this to an Info level log

that way, if need be, we can look for these message in the results watcher log gathered in the artifacts from his e2e tests

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -410,5 +465,78 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType,
 		Err: writer,
 	}, logChan, errChanRepeater)
 
+	logger.Debugw("Exiting streaming logs",
```
gabemontero, line 468: while @pmacik is testing for us, let's temporarily change this to an Info level log

that way, if need be, we can look for these message in the results watcher log gathered in the artifacts from his e2e tests

gabemontero (changes requested):
have more requests for changes

pkg/watcher/reconciler/dynamic/dynamic.go
ramessesii2: done

pkg/watcher/reconciler/dynamic/dynamic.go
ramessesii2: The reason why I'm deferring close(errChanRepeater) in this goroutine is because it's important to tell the receiving goroutines that all data have been sent which is here:
```
tknlog.NewWriter(logType, true).Write(&cli.Stream{
		Out: writer,
		Err: writer,
	}, logChan, errChanRepeater)
```
else the above code will not exit the execution. I confirmed this as well while debugging.

pkg/watcher/reconciler/dynamic/dynamic.go
ramessesii2: done, I've done a bit more of refactoring. Hopefully, this is what you're asking as well here. Pushing the changes now.

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -388,19 +397,65 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType,
...
+
+	go func(ctx context.Context, echan <-chan error, o metav1.Object) {
+		defer close(errChanRepeater)
+		select {
+		case <-ctx.Done():
+			logger.Warnw("Context done streaming log",
+				zap.String("namespace", o.GetNamespace()),
+				zap.String("name", o.GetName()),
+			)
+		case writeErr := <-echan:
```
gabemontero, line 415: to close the loop @ramessesii2 on you comment https://github.com/tektoncd/results/pull/703/files#r1486049142 if you are only closing the channel in the go routine, what unblocks this call?

are we always hitting the `case <-ctx.Done()` path ?

ramessesii2: Hey Gabe, 
We don't hit ctx.Done() most of the time. Actually, I've never encountered that bcs of 10 min of timeout.

> case writeErr := <-echan, what unblocks this call?

My understanding is based on the fact that even if echan is closed by tkn reader, reading a closed channel, it will always be successful, returning the zero value.
We could run into problem if echan is a nil. Reading from or writing to a nil channel causes code to hang forever. But since the check was not there in the original Results code I assume we expect a non-nil channel from tknReader.
Although, in the next iteration, I'll address it.  I missed this edge case.

gabemontero: neither the tekton reader or writer are closing this channel based on the code I've seen @ramessesii2 

can you provide the code link other than this goroutine where this channel is being closed?

gabemontero: nevermind I found it :-)

gabemontero: I think it is 

https://github.com/tektoncd/results/blob/main/vendor/github.com/tektoncd/cli/pkg/log/pipeline_reader.go#L50-L51

pkg/watcher/reconciler/dynamic/dynamic.go
gabemontero: yep we are good here per the thread below

pkg/watcher/reconciler/dynamic/dynamic.go
ramessesii2: sure, the flow is tknReader.read() -> readPipelineLog() -> finally landing to readAvailablePipelineLogs() for our case since we only stream longs once PipelineRuns/TRs have a specific closed state -> [close echan (errC)](https://github.com/tektoncd/results/blob/main/vendor/github.com/tektoncd/cli/pkg/log/pipeline_reader.go#L51)

gabemontero: not sure how the original echan could ever  be nil, but in general defensive programming is good :-) 

I'll wait for you update and then we can go from there - thanks

ramessesii2:
hi @gabemontero @adambkaplan with finalizers, at least locally, I've been able to fix the race condition. There's a small caveat with finalizer. Bcs we [add finalizer](https://docs.google.com/document/d/1wswcwmw8y7BRuriHAo4-G6dlZdPTuUkPRvQm9Dxvxik/edit) for the PR to not get pruned until the streaming/sending of logs is done. Finalizer along with the PipelineRun object is stored as well.
While that might not be a deal breaker (I'm not sure though), but I find it simpler to utilize [this field](https://github.com/tektoncd/results/pull/704/files#diff-83680fa39b0bf1807ef48e2ec46f89b5ef89abf69a628ba8fac9a2e01554e473R54) from LogStatus to simply hold on to pruning until we find `isStored: true`.

gabemontero:
> hi @gabemontero @adambkaplan with finalizers, at least locally, I've been able to fix the race condition. There's a small caveat with finalizer. Bcs we [add finalizer](https://docs.google.com/document/d/1wswcwmw8y7BRuriHAo4-G6dlZdPTuUkPRvQm9Dxvxik/edit) for the PR to not get pruned until the streaming/sending of logs is done. Finalizer along with the PipelineRun object is stored as well. While that might not be a deal breaker (I'm not sure though), but I find it simpler to utilize [this field](https://github.com/tektoncd/results/pull/704/files#diff-83680fa39b0bf1807ef48e2ec46f89b5ef89abf69a628ba8fac9a2e01554e473R54) from LogStatus to simply hold on to pruning until we find `isStored: true`.

for me at least I like using IsStored() instead as well @ramessesii2 @adambkaplan @ramessesii2 @sayan-biswas  assuming the watcher reconciler that handles pruning can watch for that and requeue if it is not yet stored

I believe that is the case based on what I recall from the fix for handling cancelled pipeline/task runs

putting metadata i.e. in the object we are storing seems more fragile to me in hindsight

what do you all think?

perhaps as part of breaking out the mem leak fix from this one, either create a separate PR or a separate commit in this PR so we can compare IsStored vs. finalizer

ramessesii2:
FYI : #713 uses Logs API to address race condition

PR #704, opened 2024-02-05
fix: add stored status explicitly for logs

avinal:
# Changes

- added IsStored to LogStatus
- update the API Version to v1alpha3
- v1alpha2 will be removed later
- this will serve as a clear indication of if the logs have been stored, partially stored or not stored at all.
- this can be used to mitigate the race condition between pruning of runs and log storage.

Signed-off-by: Avinal Kumar <avinal@redhat.com>

# Submitter Checklist

# Release Notes

```release-note
action required: - the log status contains an extra field called `is_stored` to denote if the logs have been correctly stored or not
- Breaking Change: API Version is updated to v1alpha3 from v1alpha2 
```

### DIFF
--- a/pkg/api/server/v1alpha2/log/log_test.go
+++ b/pkg/api/server/v1alpha2/log/log_test.go
@@ -10,14 +10,14 @@ import (
 
 	"github.com/tektoncd/results/pkg/api/server/config"
 	"github.com/tektoncd/results/pkg/api/server/db"
-	"github.com/tektoncd/results/pkg/apis/v1alpha2"
+	"github.com/tektoncd/results/pkg/apis/v1alpha3"
 	"github.com/tektoncd/results/pkg/internal/jsonutil"
 	pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestFilePath(t *testing.T) {
-	log := &v1alpha2.Log{
+	log := &v1alpha3.Log{
 		ObjectMeta: v1.ObjectMeta{
 			Name:      "test-log",
 			Namespace: "test",
@@ -112,13 +112,13 @@ func TestParseName(t *testing.T) {
 }
 
 func TestToStorage(t *testing.T) {
-	log := &v1alpha2.Log{
+	log := &v1alpha3.Log{
 		ObjectMeta: v1.ObjectMeta{
 			Name: "test-taskrun-log",
 		},
-		Spec: v1alpha2.LogSpec{
-			Type: v1alpha2.FileLogType,
-			Resource: v1alpha2.Resource{
+		Spec: v1alpha3.LogSpec{
+			Type: v1alpha3.FileLogType,
+			Resource: v1alpha3.Resource{
 				Name: "test-taskrun",
 			},
 		},
@@ -133,7 +133,7 @@ func TestToStorage(t *testing.T) {
 	rec := &pb.Record{
 		Name: "test-log",
 		Data: &pb.Any{
-			Type:  v1alpha2.LogRecordType,
+			Type:  v1alpha3.LogRecordType,
 			Value: want,
 		},
 	}
@@ -183,16 +183,16 @@ func TestToStream(t *testing.T) {
 				ResultName: "push-main",
 				Name:       "taskrun-compile-log",
 				ID:         "a",
-				Type:       v1alpha2.LogRecordType,
-				Data: jsonutil.AnyBytes(t, &v1alpha2.Log{
+				Type:       v1alpha3.LogRecordType,
+				Data: jsonutil.AnyBytes(t, &v1alpha3.Log{
 					ObjectMeta: v1.ObjectMeta{
 						Name:      "test-log",
 						Namespace: "test",
 						UID:       "test-uid",
 					},
-					Spec: v1alpha2.LogSpec{
-						Type: v1alpha2.FileLogType,
-						Resource: v1alpha2.Resource{
+					Spec: v1alpha3.LogSpec{
+						Type: v1alpha3.FileLogType,
+						Resource: v1alpha3.Resource{
 							Namespace: "app",
 							Name:      "taskrun-compile",
 						},
@@ -200,7 +200,7 @@ func TestToStream(t *testing.T) {
 				}),
 			},
 			want: &mockStream{
-				streamType: string(v1alpha2.FileLogType),
+				streamType: string(v1alpha3.FileLogType),
 			},
 		},
 		{
--- a/pkg/api/server/v1alpha2/logs.go
+++ b/pkg/api/server/v1alpha2/logs.go
@@ -21,7 +21,7 @@ import (
 	"github.com/tektoncd/results/pkg/api/server/v1alpha2/auth"
 	"github.com/tektoncd/results/pkg/api/server/v1alpha2/log"
 	"github.com/tektoncd/results/pkg/api/server/v1alpha2/record"
-	"github.com/tektoncd/results/pkg/apis/v1alpha2"
+	"github.com/tektoncd/results/pkg/apis/v1alpha3"
 	"github.com/tektoncd/results/pkg/logs"
 	pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -47,7 +47,7 @@ func (s *Server) GetLog(req *pb.GetLogRequest, srv pb.Logs_GetLogServer) error {
 		return err
 	}
 	// Check if the input record is referenced in any logs record in the result
-	if rec.Type != v1alpha2.LogRecordType {
+	if rec.Type != v1alpha3.LogRecordType {
 		rec, err = getLogRecord(s.db, parent, res, name)
 		if err != nil {
 			s.logger.Error(err)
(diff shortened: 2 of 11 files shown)


pkg/apis/v1alpha2/types.go
```diff
@@ -49,8 +49,9 @@ const (
 
 // LogStatus defines the current status of the log resource.
 type LogStatus struct {
-	Path string `json:"path,omitempty"`
-	Size int64  `json:"size"`
+	Path     string `json:"path,omitempty"`
+	Size     int64  `json:"size"`
+	IsStored bool   `json:"is_stored"`
```
sayan-biswas, line 54: `isStored` is more common than snake case json names.

avinal:
As discussed in the WG meeting, I have tested it for partial log store, and it seems there is only one issue (not ATM) we need to take care of. 
- In case of file based storage if the logs have been partially stored, we need to have a mechanism to clear them because the server will try to store the logs again, and those partially stored ones would be orphaned. we can also develop another mechanism to resume from where it broke.

In case of S3 or GCS the partial logs will be discarded.

avinal:
cc @khrm @sayan-biswas @ramessesii2 @enarha

sayan-biswas:
Please also keep in mind this is an API change, previously stored data will not have the field `isStore` and with the current design `GetLog` will fail even it the log is there.

ramessesii2:
It seems we need to add WIP tag to this PR if it's still being worked on. Thanks.

avinal:
I think v1alpha2 should be removed in a different PR or commit, please let me know if I should do that.

pkg/api/server/v1alpha3/logs.go
```diff
@@ -60,9 +60,19 @@ func (s *Server) GetLog(req *pb.GetLogRequest, srv pb.Logs_GetLogServer) error {
...
+
+	// Handle v1alpha2 and v1alpha3 differently until v1alpha2 is deprecated
+	if "results.tekton.dev/v1alpha3" == object.APIVersion {
+		if !object.Status.IsStored || object.Status.Size == 0 {
+			s.logger.Errorf("no logs exist for %s", req.GetName())
+			return status.Error(codes.NotFound, "Log doesn't exist")
+		}
+	} else {
+
+		if object.Status.Size == 0 {
```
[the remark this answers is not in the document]
avinal: You are correct, I will update the comment.

gabemontero:
> Why are we adding stored status for logs?

It is in relation to fixing https://github.com/tektoncd/results/issues/514 @khrm as the use of labels or annotations had undesirable side effects, but that should be explicitly stated in the PR's description, which as I look now, I don't see it.

Good catch @khrm

pkg/api/server/v1alpha3/logs.go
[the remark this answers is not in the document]
gabemontero: comment looks good

avinal:
@khrm @gabemontero I have broken the commits in two and also updated the PR description.

ramessesii2 (commented):
lgtm

cmd/api/main.go
```diff
@@ -180,19 +180,19 @@ func main() {
 			recovery.StreamServerInterceptor(recovery.WithRecoveryHandler(recoveryHandler)),
 		),
 	)
-	v1alpha2pb.RegisterResultsServer(gs, v1a2)
+	v1alpha3pb.RegisterResultsServer(gs, v1a2)
```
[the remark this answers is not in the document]
avinal: I did the tests, but I think something is getting wrong. If I update the server with the new code on the same DB, the logs are not being stored. I will try to find out the reason and fix.

khrm (commented):
Does this change require proto change? We are changing version of Log version in DB.
Just changing https://github.com/tektoncd/results/blob/84acda76c4cc0193eaa63d443148e912d2c192ba/pkg/apis/v1alpha3/types.go should work. 
Should TaskRunLog stored in DB mirror GRPC version?

cmd/api/main.go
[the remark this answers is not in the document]
khrm: We change the API version for breaking change only. Otherwise, you need to add proto for both v1alpha2 and v1alpha3.  Adding a field or method isn't a breaking change. Renaming or removing them is a breaking change.

In this PR, I don't find any new field or method in proto. Only an internal behaviour change which I don't think requires a breaking change.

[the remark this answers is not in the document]
avinal: Should we roll back to just migrating the DB?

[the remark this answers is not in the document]
khrm: Why do we need to migrate the DB? Isn't it part of json field?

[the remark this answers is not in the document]
avinal: Yes but existing data will not have this field. Although the condition is handled in getLog logic. If that is not a concern then we may not bump the API version. 

But then how do we indicate the change in Log object? Or is it not needed?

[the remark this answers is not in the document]
khrm: Why don't we update the Log Object version? Why is it necessary for it to have the same version as API?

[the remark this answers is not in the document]
avinal: It is not, we discussed in the call that to keep things in sync we can update the whole API. Do you think we should rediscuss the context and implementation for this in the next call?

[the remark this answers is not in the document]
khrm: Yes, we should rediscuss this.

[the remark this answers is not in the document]
gabemontero: adding object specific versioning is atypical at least from a k8s perspective

is there precedent in tekton @khrm that I'm missing?

that said @avinal if you and @sayan-biswas have not had a chance to get together and sort out the issues in your testing that we discussed, at this point I'm up for any path with is both a) not too crazy, and b) does not require a migrator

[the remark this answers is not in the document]
khrm: We are updating the Object version here as well as the API. That object is internal to Results. That was the context of my comment. Let's discuss this in WG call.

pkg/api/server/v1alpha2/logs.go
```diff
@@ -60,9 +60,20 @@ func (s *Server) GetLog(req *pb.GetLogRequest, srv pb.Logs_GetLogServer) error {
 		s.logger.Error(err)
 		return status.Error(codes.Internal, "Error streaming log")
 	}
-	if object.Status.Size == 0 {
-		s.logger.Errorf("no logs exist for %s", req.GetName())
-		return status.Error(codes.NotFound, "Log doesn't exist")
+
+	// Handle v1alpha3 and v1alpha3 differently until v1alpha3 is deprecated
```
gabemontero, line 64:
```suggestion
	// Handle v1alpha2 and earlier differently from v1alpha3 until v1alpha2 and earlier are deprecated
```

pkg/api/server/v1alpha2/log/log_test.go
```diff
@@ -112,13 +112,13 @@ func TestParseName(t *testing.T) {
 }
 
 func TestToStorage(t *testing.T) {
-	log := &v1alpha2.Log{
+	log := &v1alpha3.Log{
```
gabemontero, line 115: should we not have at least one v1alpha2 test case to validate that behavior?

pkg/api/server/v1alpha2/logs_test.go
```diff
@@ -128,19 +128,20 @@ func TestGetLog(t *testing.T) {
 		Record: &pb.Record{
 			Name: record.FormatName(res.GetName(), "baz"),
 			Data: &pb.Any{
-				Type: v1alpha2.LogRecordType,
-				Value: jsonutil.AnyBytes(t, &v1alpha2.Log{
-					Spec: v1alpha2.LogSpec{
-						Resource: v1alpha2.Resource{
+				Type: v1alpha3.LogRecordType,
```
gabemontero, line 131: shouldn't we have at least one v1alpha2 test case to validate that behavior?

pkg/watcher/convert/convert_test.go
```diff
@@ -320,7 +320,7 @@ func toJSON(v any) []byte {
 }
 
 func TestToLogProto(t *testing.T) {
-	wantType := "results.tekton.dev/v1alpha2.Log"
+	wantType := "results.tekton.dev/v1alpha3.Log"
```
gabemontero, line 323: will convert.go ever encounter v1alpha2?

gabemontero:
bump on my last round of comments @avinal :  https://github.com/tektoncd/results/pull/704/files/505ae4eb5ed61f87d578495f972a1729f72fba62

gabemontero:
also @avinal - just noticed something when looking at @khrm 's event PR .... should we be bumping the annotation for logs  to v1alpha3 ?

See https://github.com/tektoncd/results/pull/748/files#diff-83680fa39b0bf1807ef48e2ec46f89b5ef89abf69a628ba8fac9a2e01554e473R12

pkg/apis/v1alpha3/types.go
```diff
@@ -49,12 +49,13 @@ const (
 
 // LogStatus defines the current status of the log resource.
 type LogStatus struct {
-	Path string `json:"path,omitempty"`
-	Size int64  `json:"size"`
+	Path     string `json:"path,omitempty"`
+	Size     int64  `json:"size"`
+	IsStored bool   `json:"isStored"`
```
gabemontero, line 54: let's add an `errorOnStoreCode int `json:"errorOnStore"` field and an `errorOnStoreMsg string `json:"errorOnStoreMsg"`  field

that way we can tell if 

- an attempt to store has not yet completed
- the store was successful
- the store failed on an error
- the precise error is retryable or not

gabemontero (changes requested):
Some additional thoughts occurred to me @avinal @sayan-biswas @khrm @enarha @ramessesii2 while I was considering this change for helping with fixing the memory leak and canceled context issues with sufficient latency / performance characteristics

let's add this to my prior ask for additional unit tests

also @avinal are you going to be able to spend time on wrapping this PR up this week?  if not, let me know and I'll see about getting this over the finish line

thanks

avinal:
> also @avinal - just noticed something when looking at @khrm 's event PR .... should we be bumping the annotation for logs to v1alpha3 ?

Yes, that was the specific change we needed to differentiate v1alpha3 logs from v1alpha2.

pkg/api/server/v1alpha2/log/log_test.go
avinal: The tests won't work because we no longer have v1alpha2, to simulate, we have to retain the v1alpha2 types as well.

pkg/apis/v1alpha3/types.go
avinal: Yes, that seems a better idea. Should I replace `IsStored` or add these new fields?

gabemontero: let's keep IsStored and add the other two

khrm: What's `errorOnStoreCode int`?

pkg/watcher/convert/convert_test.go
khrm: Watcher doesn't need v2. Only API results can encounter v2.

gabemontero: ack thanks for confirmation @khrm

pkg/apis/v1alpha3/types.go
gabemontero: GRPC has a standard set of error codes per https://grpc.github.io/grpc/core/md_doc_statuscodes.html

In fact, when you make metric queries like 
```
grpc_server_handled_total{job="tekton-results-api", grpc_service=~"tekton.results.v1alpha2.*", grpc_method='UpdateLog'}
```

you'll see a grpc_code column with each of those status codes listed

I envision potentially gating retry on log storage errors based on those error codes, similar to what we do in k8s controllers @khrm

khrm (commented):
This is already merged here: https://github.com/tektoncd/results/pull/760

PR #708, opened 2024-02-08
fix syntax error with tekton results data model diagram

cmorinupgrade:
# Changes

Mermaid does not support titles for `graph`. If we want to have a title we would need to proceed with:
```
graph BT
  title[Tekton Results Data Model]
  B(TaskRun) --> |Record| A[Result]
  C(Log) --> |Record| A
  D(PipelineRun) --> |Record| A
```
which would produce:
image

However, there is a header just above the diagram so the title is redundant.

I propose that we remove the title completely.

/kind docs

# Release Notes

```release-note
docs: Fix mermaid.js diagram for the Results data model
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -77,9 +77,6 @@ sequenceDiagram
 ## Data Model
 
 ```mermaid
----
-title: Tekton Results Data Model 
----
 graph BT
   B(TaskRun) --> |Record| A[Result]
   C(Log) --> |Record| A


AlanGreene:
Hey @cmorinupgrade, is this related to an issue on the docs site? If so it might just be a matter of updating the mermaid version used on the website as the current syntax is working on GitHub:
https://github.com/tektoncd/results?tab=readme-ov-file#data-model
image

AlanGreene:
Title support was added in [mermaid 9.3.0](https://github.com/mermaid-js/mermaid/releases/tag/v9.3.0)

> feat: Title support via front matter for ER, state, class, git, and graph diagrams (https://github.com/mermaid-js/mermaid/pull/3706) @MasonM

AlanGreene:
Sounds good. I've also opened a PR to see if the updated version works on the website 🤞 https://github.com/tektoncd/website/pull/588

adambkaplan (approved):
/approve

Per @AlanGreene - this will work with the docs site as well.

I've added a release note, :crossed_fingers: the bot is happy, in which case I'll add lgtm too.

avinal:
Ah I was unable to respond in time: This issue is known and I had an issue about it, closed due to inactivity: https://github.com/tektoncd/website/issues/544. As per last discussion, we are using a certain hugo theme in our tektoncd website. Last time I checked the theme hasn't updated the mermaid version. It seems to be updated now.

PR #717, opened 2024-02-21
AIP-157 Partial response implementation

sayan-biswas:
In the present implementation, a user cannot filter the data in the response message, a feature provided by modern API implementations. This is particularly needed for the `List` APIs as these APIs return the whole object in the list. In scenario  where the list and object definition are big, there's a lot network bandwidth wasted because the response of the `List` API is mostly used to display a list of resources and the `Get` API is then used to fetch the individual resource.
This feature provides an easy and basic way to filter the response message across all APIs, which reduces the size of the payload and hence allows faster transfer over network.

# Changes
This feature add capabilities to filter the response message from across all APIs.
AIP detail:
https://google.aip.dev/157

# Submitter Checklist

# Release Notes
```release-note
API: Request message allows an additional parameter `fields` to filter the response message.
```


internal/fieldmask/fieldmask_test.go
```diff
@@ -0,0 +1,130 @@
...
+			Type:  "type-2",
+			Value: []byte(gjson.Parse(j).String()),
+		},
+	},
+}
+
+func TestNew(t *testing.T) {
+	f := &fieldmaskpb.FieldMask{Paths: p}
+	f.Normalize()
+	want := fm
```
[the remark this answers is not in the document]
gabemontero: and any clarification you add here @sayan-biswas should become comments in the code around here, explaining expectations, http refs to field mask doc, etc.

internal/fieldmask/fieldmask.go
```diff
@@ -0,0 +1,165 @@
...
+func (fm FieldMask) Filter(message proto.Message) {
+	if len(fm) == 0 {
+		return
+	}
+
+	reflect := message.ProtoReflect()
+	reflect.Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool {
+		mask, ok := fm[string(fd.Name())]
+		if !ok {
+			reflect.Clear(fd)
```
gabemontero, line 50: the godoc for `Clear` says it is not safe for concurrent use

are we 100% sure are path to here will not be concurrent?  I'm guessing it is now, assuming GRPC manages individual messages simlarly, but I'm still new to GRPC, so don't what to assume.

internal/fieldmask/fieldmask.go
```diff
@@ -0,0 +1,165 @@
...
+				return true
+			})
+		case fd.IsList():
+			list := reflect.Get(fd).List()
+			for i := 0; i < list.Len(); i++ {
+				mask.Filter(list.Get(i).Message().Interface())
+			}
+		case fd.Kind() == protoreflect.MessageKind:
+			mask.Filter(reflect.Get(fd).Message().Interface())
+		case fd.Kind() == protoreflect.BytesKind:
```
gabemontero, line 77: so it appears `BytesKind` is where the recursion will short circuit and return, and that map,list,message cases will all eventually boil down to bytes kind, correct?

internal/fieldmask/fieldmask.go
```diff
@@ -0,0 +1,165 @@
...
+			if b := v.Bytes(); gjson.ValidBytes(b) {
+				b, err := jsoniter.Marshal(mask.FilterJSON(b, []string{}))
+				if err == nil {
+					reflect.Set(fd, protoreflect.ValueOfBytes(b))
+				}
+			}
+		}
```
gabemontero, line 84: feels like we should have a default case to at least log unexpected messages, no?

cmd/api/main.go
```diff
@@ -169,6 +171,7 @@ func main() {
 			grpc_zap.UnaryServerInterceptor(grpcLogger, zapOpts...),
 			grpc_auth.UnaryServerInterceptor(determineAuth),
 			prometheus.UnaryServerInterceptor,
+			fieldmask.UnaryServerInterceptor(),
```
gabemontero, line 174: shouldn't we have a feature enablement check of some sort ?  presumably a client could send the extra headers or encoded data and it is OK to ignore ?

internal/fieldmask/fieldmask.go
```diff
@@ -0,0 +1,165 @@
...
+const key = "fields"
+
+// FieldMask is recursive structure to define a path mask
+type FieldMask map[string]FieldMask
+
+// New create a FieldMask from the input array of paths. The array should contain JSON paths with dit "." notation.
+func New(paths []string) FieldMask {
+	mask := make(FieldMask)
+	for _, path := range paths {
+		current := mask
```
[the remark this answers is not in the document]
gabemontero: not a blocker comment per se, but curious what you think?

gabemontero (commented):
finally got around to taking a pass at this @sayan-biswas 

of course, being brand new / a novice to all this, consider that perspective when reviewing my feedback ;-)

internal/fieldmask/fieldmask.go
sayan-biswas: The filtering happens in the response proto. There can be only one response to a request, handled by a single goroutine. If in case we are modifying the response in another Interceptor, still I guess it won't be concurrent access because the interceptors are chained.

internal/fieldmask/fieldmask.go
sayan-biswas: In our implementation, most of the cases will boil down to `BytesKind`, but not all. In some messages, there are other types of fields also. But this is more of a generic implementation, in case in future we include child proto or `Lists` in our messages.

internal/fieldmask/fieldmask.go
sayan-biswas: I would say no, because this is simply filtering. Any unknown cases will just be ignored.
And for logging, there's are some restrictions here. We cannot throw `error` because then the client will receive the  error. We cannot bring logger from the context, then the implementation will have dependency and can't be used as a library.
Although we can still log with normal golang's `fmt` or `logger` library, but that will break structured logging.

internal/fieldmask/fieldmask.go
gabemontero: those clarifications help thanks

internal/fieldmask/fieldmask.go
gabemontero: valid reasoning 

I still wonder about being able to debug in the future "why did my filter not catch 'X'"

could we emit events for the errors in this switch?

could we emit events for the default case?  Or are there too many types besides MessageKind and BytesKind?

internal/fieldmask/fieldmask.go
gabemontero: ack thanks for confirming

cmd/api/main.go
sayan-biswas: I have added a feature flag implementation to handle all features control.

internal/fieldmask/fieldmask_test.go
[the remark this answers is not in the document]
sayan-biswas: `a.b.(c)` is not required because the `c` is child of `b`. So if the `b` is included in the field mask then the response will include all the child and its recursive child in the response.

internal/fieldmask/fieldmask.go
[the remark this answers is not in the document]
sayan-biswas: I have made this recursive, but changed the name of the method to `Build`, since `New` almost always represent a non-receiver method in most programming language. And in this case we do need a receiver to keep track of the current path. 
I am not sure if this could have been converted to recursive without an extra variable.

internal/fieldmask/fieldmask.go
```diff
@@ -0,0 +1,165 @@
...
+		fields := strings.Split(path, ".")
```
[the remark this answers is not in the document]
sayan-biswas: Added reference and an example in godoc format.

internal/fieldmask/fieldmask_test.go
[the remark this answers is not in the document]
gabemontero: thanks @sayan-biswas that helps ..... in hindsight it is obvious, just did occur to me during my last review  :-)

internal/fieldmask/fieldmask.go
gabemontero: yes k8s events .... api server has a k8s client so in theory could emit events to help with debug, and you don't have the logging pain you noted earlier

internal/fieldmask/fieldmask.go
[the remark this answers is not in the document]
gabemontero: awesome thanks

internal/fieldmask/fieldmask.go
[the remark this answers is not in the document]
gabemontero: regardless of that last poing @sayan-biswas (fwiw I don't see another way other than what you did) I don't know what you think, but this version is way more elegant and understandable.

works for me - thanks

config/base/env/config
```diff
@@ -6,13 +6,7 @@ DB_NAME=tekton-results
 DB_SSLMODE=disable
 DB_SSLROOTCERT=
 DB_ENABLE_AUTO_MIGRATION=true
-DB_MAX_IDLE_CONNECTIONS=10
```
gabemontero, line 9: I'm assuming you did this @sayan-biswas because I didn't need to set these defaults when I added the tuning options ?  If so, thanks for the correction.

gabemontero: Or was this accidental?  ...If it was, in hindsight, I does probably make sense to not set these, even to the defaults. (though 10 does not line up with the default as described in the doc.

Users can always add these to their config map.

cmd/api/README.md
```diff
@@ -2,42 +2,36 @@
...
-| DB_ENABLE_AUTO_MIGRATION | Auto-migrate the database on startup (create/update schemas). For further details, refer to <https://gorm.io/docs/migration.html> | true (default)                                                                               |
-| PROFILING                | Enable profiling server                                                                                                           | false  (default)                                                                             |
-| PROFILING_PORT           | Profiling Server Port                                                                                                             | 6060  (default)                                                                              |
-| DB_MAX_IDLE_CONNECTIONS  | The number of idle database connections to keep open                                                                              | 2 (default for golang, but specific database drivers may have settings for this too)         |
```
gabemontero, line 16: You deleted all my tuning options here in the README @sayan-biswas .... we need to minially document them, even if they are not specified in the config/base/env/config file

cmd/api/main.go
gabemontero: feature gate looks good @sayan-biswas good think .... in looking at it though I raised some questions around changes you made to my recently added tuning options 

I'll submit a separate review for those

gabemontero (changes requested):
just a doc bit I think @sayan-biswas

gabemontero:
I don't seem to have permission to resolve comment threads @sayan-biswas 

Feel free to resolve all of mine except https://github.com/tektoncd/results/pull/717#pullrequestreview-2038043891

internal/fieldmask/fieldmask.go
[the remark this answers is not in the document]
sayan-biswas: Agree.

gabemontero:
bump @sayan-biswas on my last round of comments - https://github.com/tektoncd/results/pull/717#pullrequestreview-2038043891

cmd/api/README.md
sayan-biswas: Ohh.. Not sure why how this was missed. I'll rebase again to include the changes.

khrm (commented):
@sayan-biswas Can you rebase it again without removing Gabe's performance tuning changes?

sayan-biswas:
> @sayan-biswas Can you rebase it again without removing Gabe's performance tuning changes?

Sorry! I was caught up with other activities. I will rebase soon.

cmd/api/README.md
```diff
@@ -42,6 +42,7 @@
 | STORAGE_EMULATOR_HOST    | GCS Storage Emulator Server                                                                                                       | http://localhost:9004                                                                        |
 | CONVERTER_ENABLE         | Whether to start converter of v1beta1 TaskRun/PipelineRun records to v1                                                           | true                                                                                         |
 | CONVERTER_DB_LIMIT       | How many records to convert at a time in a transaction                                                                            | 50                                                                                           |
+| FEATURE_GATES            | Configuration to enable/disable a feature                                                                                         | PartialResponse=true,foo=false,bar=true                                                      |
```
enarha, line 45: @vdemeester what do you think about that name `FEATURE_GATES`? Would you prefer if we have have separate options controlling each gate e.g `PARTIAL_RESPONSE`, `FOO`, `BAR`, instead of the list `PartialResponse=true,foo=false,bar=true`?

khrm:
/hold cancel

Tested this. It's working fine.

khrm (commented):
/lgtm

PR #712, opened 2024-02-14
employ some channel best practices to avoid hanging goroutines

gabemontero:
# Changes

Fixes https://github.com/tektoncd/results/issues/695

possibly

my team (including @ramessesii2 @avinal @sayan-biswas @pmacik) is doing some internal testing to both 
zero in on the cause of the leak, as well as see if our theory around the goroutines not getting GC'ed based on what we 
see in testing proves out.

I've marked this as WIP and hold because aside from still needing to verify things, this PR currently has some debugging prints we will eventually remove.

/kind bug
/hold

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you review them:

- [ ] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] [Tested your changes locally](https://github.com/tektoncd/results/blob/main/docs/DEVELOPMENT.md) (if this is a code change)
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Has a kind label. You can add a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user-facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contain the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Fixes around the threading / goroutines employed while storing logs mitigate an observed memory leak
```

PR #725, opened 2024-03-08
add timeout to dynamic Reconcile; switch from CloseSend to CloseRecv in streamLogs; add goroutine dump on timeouts/deadlocks

gabemontero:
# Changes

So we ran with the change from #712 in our production system, and while we confirmed the results logging memory leak was addresed, after about 12 to 13 hours, our reconciler threads systematically became deadlocked, and our watcher quit processing events.
    
We as of yet have not been able to get a goroutine dump with stack traces when this problem occurs, so we are unclear
whether the #712 fixes have had a direct cause to the deadlock, or if another issue was encountered.  Among other things our api server container restarted during the watcher deadlock, where the previous pod logs gave no clear indicationas to why.
    
This change pulls a couple of potentially helpful bits to help either diagnose or work around the deadlock:
    1) we have added a timeout to the context used in the dynamic Reconcile method, in case a blockage in any RPC call using a context somehow was causing the problem
    2) we also employ the combination of cancelling the context on method exit, to again unblock things, as well as the switch to
    CloseAndRecv instead of CloseSend to confirm the UpdateLog finished, so that our canceling of the streamLog context does not
    intermittenly cancel an UpdateLog call that would have otherwise succeeded.
    3) we are analyzing how a context is released, and if it is from a timeout and not cancel, we initiate a goroutine dump with stack traces
   4) using of a context with timeout that is canceled on exit from the reconcile method require no longer running 'sendLogs' on a separate goroutine, otherwise we re-introduced intermitent cancelling of 'UpdateLog' processing before it could complete.
   5) we now log the dealines for UpdateLog on the api server side 

rh-pre-commit.version: 2.2.0
rh-pre-commit.check-secrets: ENABLED

/kind bug

# Submitter Checklist

- [n/a ] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [n/a ] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ y] [Tested your changes locally](https://github.com/tektoncd/results/blob/main/docs/DEVELOPMENT.md) (if this is a code change)
- [y ] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ y] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [y ] Has a kind label. You can add a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [ y] Release notes block below has been updated with any user-facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [n/a ] Release notes contain the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Address deadlocked reconciler threads potentially hung while streaming logs
```

### DIFF
--- a/pkg/watcher/reconciler/dynamic/dynamic.go
+++ b/pkg/watcher/reconciler/dynamic/dynamic.go
@@ -18,6 +18,9 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"os"
+	"runtime/pprof"
+	"strings"
 	"time"
 
 	"github.com/fatih/color"
@@ -36,6 +39,8 @@ import (
 	"github.com/tektoncd/results/pkg/watcher/results"
 	pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
 	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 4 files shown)


gabemontero:
OK https://github.com/tektoncd/results/pull/725/commits/4cc4cab884d948928125ef7021213bb2a82aa2ee passed the e2e's

Next will in a separate commit attempt a more expansive use of context with timeout around all grpc calls from the dynamic reconciler.

gabemontero:
ok @sayan-biswas I have 2 forms of a fix and/or work around for the results watcher hang we saw in our prod env today

the first commit just covers updating logs to the results api server

the second commit handles all grpc calls to the api  server coming out of the dynamic reconciler

after some code review, it centers on the notion that we were blocked on an grpc call to the api server, where the api server may have been in a bad state, given I noticed an api restart around the time I recycled the watcher pod

So basically,

1. by setting a timeout on the context used for all grpc calls out of the dynamic reconciler, we timeout and exit after 5 minutes if any of those calls are hung
2. the second commit, in order to have 1 context cover both the log grpc and non-log gprc, has to move the call to streamLog back on the reconciler thread, as cancelling/cleaning up the parent context could inadvertently abort the UpdateLog call
3. also employed the `CloseAndRecv` improvement on the UpdateLog call to again coordinate with cleaning up the context

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -333,30 +336,31 @@ func (r *Reconciler) sendLog(ctx context.Context, o results.Object) error {
...
+
 	}
 
 	return nil
 }
 
 func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType, logName string) error {
-	logger := logging.FromContext(ctx)
-	logsClient, err := r.resultsClient.UpdateLog(ctx)
+	streamCtx, streamCancel := context.WithTimeout(ctx, 5*time.Minute)
```
gabemontero, line 360: if we go with the broader using of context timeout within the dynamic reconciler loop, we would get rid of this particular child context with timeout

gabemontero:
ok @sayan-biswas I now have a third commit that deciphers if the context or any specific GRPC calls exist because of deadline exceeded, and if so, dumps a goroutine with stack traces list to stdout for analysis.

So, with our hand on Friday, either the tkn client threading ended in a deadlock, or more likely, we had a grpc call to the api server hang (perhaps because the api server became messed up, which might have occurred since it restarted on its own).

In either case, we should get a thread dump in the log that we can diagnose.

gabemontero:
e2e's green with timeout full propagated, no go func for sendLogs

will update configuration of timeout now

gabemontero:
@sayan-biswas @khrm @avinal @enarha this PR is ready for review, iterate, merge

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -89,14 +94,85 @@ func NewDynamicReconciler(rc pb.ResultsClient, lc pb.LogsClient, oc ObjectClient
...
 // Reconcile handles result/record uploading for the given Run object.
 // If enabled, the object may be deleted upon successful result upload.
 func (r *Reconciler) Reconcile(ctx context.Context, o results.Object) error {
-	logger := logging.FromContext(ctx)
+	var dynamicContext context.Context
+	var dynamicCancel context.CancelFunc
+	// context with timeout does not work with the partial end to end flow that exists with unit tests;
+	// this field will alway be set for real
+	if r.cfg != nil && r.cfg.UpdateLogTimeout != nil {
+		dynamicContext, dynamicCancel = context.WithTimeout(ctx, *r.cfg.UpdateLogTimeout)
```
khrm, line 128:
```suggestion
		ctx, dynamicCancel = context.WithTimeout(ctx, *r.cfg.UpdateLogTimeout)
```

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -333,22 +414,21 @@ func (r *Reconciler) sendLog(ctx context.Context, o results.Object) error {
 			zap.String("name", o.GetName()),
 		)
 
-		go func() {
```
khrm, line 336: Why don't we pass context here and use select and ctx.done?

gabemontero: because the end result is the same; we block the reconcile thread until the operation is complete

no reason for the added complextiy

pkg/watcher/reconciler/dynamic/dynamic.go
gabemontero: yeah I'm not a fan of using the generic variable name for this ... I was intentional in this name change

can you elaborate @khrm on why you want the name to be `ctx` ?

pkg/watcher/reconciler/dynamic/dynamic.go
```diff
@@ -396,6 +499,13 @@ func (r *Reconciler) streamLogs(ctx context.Context, o results.Object, logType,
 		Err: inMemWriteBufferStderr,
 	}, logChan, errChan)
 
+	// pull the first error that occurred and return on that; reminder - per https://golang.org/ref/spec#Channel_types
+	// channels act as FIFO queues
+	chanErr, ok := <-errChan
```
gabemontero, line 504: since @khrm last review I've added here a simplified version of the error handling that @sayan-biswas used to do prior to pr #712 

let's just return the first error seen; from what I see in the tkn code, multiple errors are not collected

with this though @sayan-biswas and the prior changes made with #712, plus moving the `sendLogs` call back on the reconciler thread, we now address one of the short comings of the old implementation, namely ignoring of errors and preventing retries on reconciliation.

Reminder:  with the `threadiness` arg, users can adjust the number of threads when they enable log support.

I will add a separate PR to add the k8s client tuning shortly

pkg/watcher/reconciler/dynamic/dynamic.go
khrm: Someone modifying the function later on can easily by mistake use `ctx`. `ctx` is still available.

gabemontero: I've pushed as a separate commit for now @khrm to facilitate your review  .... I can squash once we are sync'ed on the change being ready

pkg/watcher/reconciler/dynamic/dynamic.go
gabemontero: https://github.com/tektoncd/results/pull/744 is the add tuning options PR

gabemontero:
@khrm @sayan-biswas this is ready for review again .... once we are good, I can squash as needed .. thanks

khrm (commented):
/approve

sayan-biswas (approved):

sayan-biswas:
> @khrm @sayan-biswas this is ready for review again .... once we are good, I can squash as needed .. thanks

Changes looks fine.
Please go ahead and squash.

gabemontero:
thanks @sayan-biswas @khrm commits are squashed

PR #740, opened 2024-03-24
Don't run auth check for every log fragment

stuartwdouglas:
Currently when receiving logs the auth check is performed on every log fragment. Given that there is already a check that the name is the same, and the auth check only depends on the name, we should be able to skip it after the first one.


khrm (commented):
/lgtm

avinal (approved):
Works as expected
/approve

PR #744, opened 2024-04-01
Add ability to set DB connection pool limits, grpc worker pool size, k8s client qps/burst for both api server and watcher

gabemontero:
# Changes

pulls in https://github.com/tektoncd/results/pull/741 plus adds
- grpc worker pool size control
- k8s qps/burst control for both api server and watcher

/kind feature

# Submitter Checklist

- [/ ] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ n/a] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ /] [Tested your changes locally](https://github.com/tektoncd/results/blob/main/docs/DEVELOPMENT.md) (if this is a code change)
- [ /] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ /] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [ /] Has a kind label. You can add a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [ /] Release notes block below has been updated with any user-facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ n/a] Release notes contain the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
The number of idle and max database connections, grpc worker pool goroutines, k8s client qps and burst,  in the api server can be now configured using environment variables.
The k8s client qps and burst settings can be set via command line arguments on the watcher
```


gabemontero:
@khrm @sayan-biswas @avinal @enarha PTAL

gabemontero:
bump @khrm @sayan-biswas can I get a look at this please ??

also, in relation to the discussions in https://github.com/tektoncd/results/pull/741 about the db pool size, if it will expedite things, I'm fine at this point setting the defaults to 2 or whatever default size will expedite this getting merged

thanks

khrm:
@gabemontero Sure. I will take a look tomorrow.

BTW, I wanted to test the affect of db maxPool to 10.

cmd/watcher/main.go
```diff
@@ -76,6 +79,16 @@ func main() {
 	// controller's workqueue.
 	controller.DefaultThreadsPerController = *threadiness
 
+	// This parses flags.
+	k8scfg := injection.ParseAndGetRESTConfigOrDie()
+
+	if qps != nil {
+		k8scfg.QPS = float32(*qps)
```
khrm, line 86: Should we multiply by the number of controllers we have? @vdemeester, we do something similar in pipelines.

khrm: So see here: https://github.com/knative/pkg/blob/5d4af76051e4daa445151321f4a53953d2ce7cbb/injection/sharedmain/main.go#L241
For default, we multiply by the number of controllers. So we should multiply for values provided also.

gabemontero: ah interesting ... and yeah by comparison in core tekton there is this FIXME from @vdemeester :-) - https://github.com/tektoncd/pipeline/blob/main/cmd/controller/main.go#L72-L75

so we have 2 controller at the top of the watcher and my change here switches to `MainWithConfig` per https://github.com/tektoncd/results/blob/89b591d63b025a0e7daecaff94034b32c243e6cf/cmd/watcher/main.go#L123-L129

and that there ^^  @khrm where we are now calling the `MainWithConfig` function you highlighted with your link https://github.com/knative/pkg/blob/5d4af76051e4daa445151321f4a53953d2ce7cbb/injection/sharedmain/main.go#L241 where we are already passing non-zero value for QPS and Burst means the code you highlighted, which I am now calling with this change, is doing the multiplication you are asking for.

So we are good here.  Correct?

khrm: That code does multiplication for zero value. To have consistent behaviour across Tekton components, we should do the multiplication.
```
	if cfg.Burst == 0 {
		cfg.Burst = len(ctors) * rest.DefaultBurst
	}
```

khrm: Ideally, knative/pkg itself should do the multiplication but I don't think that's coming anytime soon.

gabemontero: yep looked too fast your right ... will update

gabemontero: updated @khrm in separate commit to ease review

cmd/api/main.go
```diff
@@ -119,6 +122,27 @@ func main() {
 		log.Fatalf("Failed to connect to database: %v", err)
 	}
 
+	sqlDB, err := db.DB()
+
+	// Set DB connection limits
+	maxIdle := serverConfig.DB_MAX_IDLE_CONNECTIONS
+	maxOpen := serverConfig.DB_MAX_OPEN_CONNECTIONS
+	if maxOpen < 1 {
```
khrm, line 130: There's one more problem with this code is that elides standard library behavior. There's no way to get standard library behavior.

gabemontero: Not sure I follow you point entirely.  Some clarification could help.

My best guess is that you don't want to have a default setting at all.

So that if they do not set the env vars, just do nothing, and leave the current behavior.

I'm fine with that change, if that is what you mean.

khrm:
To verify this PR behavior vs default behavior, I install local DB in the cluster as we do in  `./test/e2e/01-install.sh`.

Then I ran tests once parallelly with 10:
Main branch results:
https://gist.github.com/khrm/62a6e24eb3b7926fd8678c7d257775e3
This PR results:
https://gist.github.com/khrm/58ba044b5dc9436a91ab5554a44ed020

I merged the main branch to this PR to avoid any discrepancy from earlier PRs. We can see that the former is more performant than the latter.

cmd/api/main.go
khrm: Yes, that's what I meant. Or setting 0 means we are going to use the current behavior.

gabemontero:
> To verify this PR behavior vs default behavior, I install local DB in the cluster as we do in `./test/e2e/01-install.sh`.
> 
> Then I ran tests once parallelly with 10: Main branch results: https://gist.github.com/khrm/62a6e24eb3b7926fd8678c7d257775e3 This PR results: https://gist.github.com/khrm/58ba044b5dc9436a91ab5554a44ed020
> 
> I merged the main branch to this PR to avoid any discrepancy from earlier PRs. We can see that the former is more performant than the latter.

OK I will do the change I note in https://github.com/tektoncd/results/pull/744#discussion_r1555759184 so that the main branch behavior is not changed at all unless those env are set.

Not sure what precisely is different between your tests, my tests, and @stuartwdouglas tests but I don't think it matters at this point.

As long as we can change these settings and run tests in larger test clusters to see improvements, that is sufficient.

sayan-biswas (approved):

cmd/api/README.md
```diff
@@ -13,6 +13,11 @@
 | DB_ENABLE_AUTO_MIGRATION | Auto-migrate the database on startup (create/update schemas). For further details, refer to <https://gorm.io/docs/migration.html> | true (default)                               |
 | PROFILING                | Enable profiling server                                                                                                           | false  (default)                             |
 | PROFILING_PORT           | Profiling Server Port                                                                                                             | 6060  (default)                              |
+| DB_MAX_IDLE_CONNECTIONS  | The number of idle database connections to keep open                                                                              | 10 (default)                                 |
+| DB_MAX_OPEN_CONNECTIONS  | The maximum number of database connections, for best performance it should equal DB_MAX_IDLE_CONNECTIONS                          | 10 (default)                                 |
```
sayan-biswas, line 17: This is now the Postgres driver default, which is 100.

sayan-biswas:
/approve 

The README needs a minor change and commits squashed.

cmd/api/README.md
gabemontero: well according to golang's database/sql package it is 2:
```
const defaultMaxIdleConns = 2

func (db *DB) maxIdleConnsLocked() int {
	n := db.maxIdleCount
	switch {
	case n == 0:
		// TODO(bradfitz): ask driver, if supported, for its default preference
		return defaultMaxIdleConns
	case n < 0:
		return 0
	default:
		return n
	}
}
```

and max open is unlimited (which can also have ramifications, vs. having a fixed number) 

```
// SetMaxOpenConns sets the maximum number of open connections to the database.
//
// If MaxIdleConns is greater than 0 and the new MaxOpenConns is less than
// MaxIdleConns, then MaxIdleConns will be reduced to match the new
// MaxOpenConns limit.
//
// If n <= 0, then there is no limit on the number of open connections.
// The default is 0 (unlimited).
func (db *DB) SetMaxOpenConns(n int) {
	db.mu.Lock()
	db.maxOpen = n
	if n < 0 {
		db.maxOpen = 0
	}
	syncMaxIdle := db.maxOpen > 0 && db.maxIdleConnsLocked() > db.maxOpen
	db.mu.Unlock()
	if syncMaxIdle {
		db.SetMaxIdleConns(n)
	}
}
```

I'm not finding any postgresql specific overrides in the vendor tree to those @sayan-biswas .... for my education can you tell me where you see that 100.

Regardless, it certainly no longer is 10, and I need to update.  Thanks.

gabemontero: @sayan-biswas - per the WG meeting today with myself, @avinal and @khrm we at least agreed that the `database/sql` defaults I noted above are the ones that we should document, as there are many flavors of postgresql driver, and those different drivers will have different defaults.

I will try to elaborate on this as part of the README update I am working on.  Unless unexpectedly interrupted I should have some form of this up in an hour or so I think.

cmd/api/README.md
```diff
@@ -2,37 +2,42 @@
...
+| DB_ENABLE_AUTO_MIGRATION | Auto-migrate the database on startup (create/update schemas). For further details, refer to <https://gorm.io/docs/migration.html> | true (default)                                                                               |
+| PROFILING                | Enable profiling server                                                                                                           | false  (default)                                                                             |
+| PROFILING_PORT           | Profiling Server Port                                                                                                             | 6060  (default)                                                                              |
+| DB_MAX_IDLE_CONNECTIONS  | The number of idle database connections to keep open                                                                              | 2 (default for golang, but specific database drivers may have settings for this too)         |
```
gabemontero, line 16: the additional explanation widened the columns, and is why the git delta seem larger

use the `w=1` parameter a la https://stackoverflow.com/questions/37007300/how-to-ignore-whitespace-in-github-when-comparing if you want to filter whitespace our

gabemontero:
README updated, commits squashed, PR title / description / release notes refreshed @sayan-biswas @khrm 

I suggest viewing diffs with ignoring whitespace, i.e. https://github.com/tektoncd/results/pull/744/files?w=1 to filter out the widening of the READM md columns as I added more explanation about golang vs. DB driver defaults

khrm (commented):
/approve

Looks good.

khrm:
@sayan-biswas @avinal  Can you take a cursory look? I think it looks good. We can give lgtm here.

gabemontero:
> @sayan-biswas @avinal Can you take a cursory look? I think it looks good. We can give lgtm here.

@sayan-biswas did previously apply `/approve`  with https://github.com/tektoncd/results/pull/744#issuecomment-2048356840

But of course if either @sayan-biswas or @avinal could take another look soon that would be great.

PR #713, opened 2024-02-16
[WIP] Logs API apporach to fix race condition due to pruning in results watcher

ramessesii2:
# Changes

Fixes #514
/kind bug

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you review them:

- [ ] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] [Tested your changes locally](https://github.com/tektoncd/results/blob/main/docs/DEVELOPMENT.md) (if this is a code change)
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Has a kind label. You can add a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user-facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contain the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Free up resources (PipelineRun/TaskRun) potentially without any race conditions in pruning w.r.t to streaming logs even with no Grace Period. 
```

PR #760, opened 2024-06-06
Add log stored status explicitly

khrm:
- adding log stored stored status explictly in the Log object improves the detection for partial or no storage of logs
- it might help mitigate the race condition between pruning the runs and storing the logs.

# Changes

# Submitter Checklist

# Release Notes

### DIFF
--- a/pkg/api/server/v1alpha2/log/log.go
+++ b/pkg/api/server/v1alpha2/log/log.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/tektoncd/results/pkg/api/server/config"
 	"github.com/tektoncd/results/pkg/api/server/db"
-	"github.com/tektoncd/results/pkg/apis/v1alpha2"
+	"github.com/tektoncd/results/pkg/apis/v1alpha3"
 	pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -64,13 +64,13 @@ type Stream interface {
 //
 // NewStream may mutate the Log object's status, to provide implementation information
 // for reading and writing files.
-func NewStream(ctx context.Context, log *v1alpha2.Log, config *config.Config) (Stream, error) {
+func NewStream(ctx context.Context, log *v1alpha3.Log, config *config.Config) (Stream, error) {
 	switch log.Spec.Type {
-	case v1alpha2.FileLogType:
+	case v1alpha3.FileLogType:
 		return NewFileStream(ctx, log, config)
-	case v1alpha2.S3LogType:
+	case v1alpha3.S3LogType:
 		return NewS3Stream(ctx, log, config)
-	case v1alpha2.GCSLogType:
+	case v1alpha3.GCSLogType:
 		return NewGCSStream(ctx, log, config)
 	}
 	return nil, fmt.Errorf("log streamer type %s is not supported", log.Spec.Type)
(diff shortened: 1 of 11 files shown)


khrm (commented):
/assign @gabemontero @enarha

pkg/apis/v1alpha3/types.go
```diff
@@ -1,12 +1,13 @@
...
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
 // LogRecordType represents the API resource type for Tekton log records.
-const LogRecordType = "results.tekton.dev/v1alpha2.Log"
+const LogRecordType = "results.tekton.dev/v1alpha3.Log"
+const LogRecordTypeV2 = "results.tekton.dev/v1alpha2.Log"
```
khrm, line 10: We need this also because some of the Type stored in DB can be v2.

pkg/api/server/v1alpha2/log/log.go
```diff
@@ -100,11 +100,11 @@ func ToStorage(record *pb.Record, config *config.Config) ([]byte, error) {
 // First one is a new log streamer created by log record.
 // Second one is log API resource retrieved from log record.
 // Third argument is an error.
-func ToStream(ctx context.Context, record *db.Record, config *config.Config) (Stream, *v1alpha2.Log, error) {
-	if record.Type != v1alpha2.LogRecordType {
+func ToStream(ctx context.Context, record *db.Record, config *config.Config) (Stream, *v1alpha3.Log, error) {
+	if record.Type != v1alpha3.LogRecordType || record.Type != v1alpha3.LogRecordTypeV2 {
```
khrm, line 104: We need V2 here because it can be v2 also. So I added that,

khrm (commented):
/kind feature

khrm (commented):
/assign @gabemontero

pkg/apis/v1alpha3/types.go
```diff
@@ -49,12 +52,14 @@ const (
 
 // LogStatus defines the current status of the log resource.
 type LogStatus struct {
-	Path string `json:"path,omitempty"`
-	Size int64  `json:"size"`
+	Path            string `json:"path,omitempty"`
+	Size            int64  `json:"size"`
+	IsStored        bool   `json:"isStored"`
+	ErrorOnStoreMsg string `json:"errorOnStoreMsg"`
```
gabemontero, line 58: yeah per my comment https://github.com/tektoncd/results/pull/704/files#r1630064603 @khrm lets add an `errorOnStoreCode int` as well

@avinal @sayan-biswas fyi

pkg/apis/v1alpha3/types.go
gabemontero: ack @khrm good catch

gabemontero (changes requested):
just the one additional field on the `LogStatus` obj @khrm 

thanks

pkg/apis/v1alpha3/types.go
gabemontero: @khrm consider the old prototype for using this change over in https://github.com/tektoncd/results/pull/713

I want that call in the watcher's dynamic.go to not just check if it is stored, and if it is not, requeue to consider pruning later.

I want it to also see that if it is not stored per the update made in pkg/api/server/v1alpha2/logs.go 's `handleReturn` (where it has the original "error"), but because of an error on the api server side, to see if the particular error is one of the grpc intermittent sort of hiccups that mean we can re-attempt the log storage.

In this way, we finally address the lack of retry on error that @sayan-biswas @avinal @enarha and myself have lamented for months.

Now, if there is some flaw in the way I've connected the dots here, then let's you @khrm, myself, @avinal @sayan-biswas @enarha discuss either here on in the next WG call.

As a backup, I'm also fine with setting a boolean "IsRetrablyErrorOccurred" field. in `LogStatus`, have the handleReturn code make the decision if the error that occured in an intermittent, retryable sort of thing, and then the dynamic.go code works off of that to try the `UpdateLog` call again.

thanks

khrm: I have added a new field for this.

gabemontero: LGTM @khrm thanks

@enarha @avinal @sayan-biswas as I am not in the OWNERS file anywhere, we need reviews from you all to get this finally merged. 

PTAL

avinal (approved):
/lgtm
/approve

PR #879, opened 2024-11-10
Add Support for Blob Type (s3 and GCS)

khrm:
Added support for reading forwarded logs to s3 and GCS.
If vector forwards logs to s3 and GCS, these logs can be read using
Go CDK's blob library.
Also, if result logging mechanism is changed from v1alpha2 to v1alpha3
which doesn't support watcher, we can read old logs if they exist.

# Changes

# Submitter Checklist

# Release Notes

```release-note
Added support for reading forwarded logs to s3 and GCS.
If vector forwards logs to s3 and GCS, these logs can be read using
Go CDK's blob library.
Also, if result logging mechanism is changed from v1alpha2 to v1alpha3
which doesn't support watcher, we can read old logs if they exist.
```


khrm (commented):
/assign @enarha @vdemeester

pkg/api/server/v1alpha2/log/gcs_test.go
```diff
@@ -79,6 +79,7 @@ func NewRecordReplayClient(t *testing.T, modReq func(r *httpreplay.Recorder), po
 	if err != nil {
 		t.Fatal(err)
 	}
+	rep.IgnoreHeader("X-Goog-Gcs-Idempotency-Token")
```
enarha, line 82: why do we ignore this one?

pkg/api/server/v1alpha2/plugin/plugin_logs.go
```diff
@@ -31,13 +43,32 @@ const (
...
 
+var (
+	openBucket = func(ctx context.Context, urlString string) (*blob.Bucket, error) {
+		bucket, err := blob.OpenBucket(ctx, urlString)
+		return bucket, err
+	}
+	clean = func(bucket *blob.Bucket, logger *zap.SugaredLogger) {
+		err := bucket.Close()
+		if err != nil {
+			logger.Errorf("Got error while closing blob %s", err)
```
enarha, line 63:
```suggestion
			logger.Errorf("Got error while closing bucket %s", err)
```

pkg/api/server/v1alpha2/plugin/plugin_logs.go
```diff
@@ -250,8 +271,133 @@ func (s *LogPluginServer) getLokiLogs(writer io.Writer, parent string, rec *db.R
...
+	case typePipelineRun:
+		err := errors.New("pipelinerun not supported, please use taskrun")
+		s.logger.Error(err.Error())
+		return err
+	case typeTaskRun:
+		logRec := &db.Record{}
+		if legacy {
+			logRec, err = getLogRecord(s.db, parent, rec.ResultID, rec.Name)
+			if err != nil {
+				s.logger.Debugf("error getting legacy log record: %s", err)
```
enarha, line 312: why do we continue here?
1. If err != nil, then logRec will be nil
2. If legacy is configured which means log type to v1alpha2, why do we continue as v1alpha3?

pkg/api/server/v1alpha2/plugin/plugin_logs.go
```diff
@@ -250,8 +271,133 @@ func (s *LogPluginServer) getLokiLogs(writer io.Writer, parent string, rec *db.R
...
+			}
+			if logRec != nil && err == nil {
+				log := &v1alpha3.Log{}
+				err := json.Unmarshal(logRec.Data, log)
+				if err != nil {
+					err = fmt.Errorf("could not decode Log record: %w", err)
+					s.logger.Error(err)
```
enarha, line 319: below you call the .Error() (`s.logger.Error(err.Error())`) method on the error while logging. I do not mind either, just keep consistent.
Also in the other instances below.

pkg/api/server/v1alpha2/plugin/server.go
```diff
@@ -0,0 +1,121 @@
...
+func NewLogServer(config *config.Config, logger *zap.SugaredLogger, auth auth.Checker, db *gorm.DB) (*LogServer, error) {
+	s := &LogServer{
+		config: config,
+		logger: logger,
+		auth:   auth,
+		db:     db,
+	}
+
+	s.logger.Debugf("LOGS_TYPE: %s", strings.ToLower(s.config.LOGS_TYPE))
+	// If the logs type is not Loki, we don't need to set up the LogPluginServer
```
enarha, line 59: setLogPlugin() with also return true if the log type BlobLogType, so either the comment is outdated or the logic below.

khrm: Yes. Comment is outdated.

pkg/api/server/v1alpha2/log/gcs_test.go
khrm: This unit tests uses http replay. Earlier, when we ran this on a live cluster, GCS sdk supported this differently. Now, this header gives us error, to resolve this either we ignore this header or run against live GCS bucket and regenerate test data.

pkg/api/server/v1alpha2/plugin/server.go
khrm: Updated this.

pkg/api/server/v1alpha2/plugin/plugin_logs.go
khrm: Changed it to use `s.logger.Error(err)` to everywhere.

pkg/api/server/v1alpha2/plugin/plugin_logs.go
khrm: 1. Yes, fixed that.
2. Yes, Log Type kind itself can be v1alpha2 or  v1alpha3. 
In plugin API, there's no LogType record itself.

PR #896, opened 2024-12-03
Fix the blob for s3 prefixed key 

khrm:
# Changes
Prefix behave differently than normal folders. Folders don't
require a leading "/" but Prefix does.
Also, refactor to avoid the error during the clean func call.

Also, legacy APIs should only be supported for v1beta1 or using LogType record.
This adds it to a v1beta1 Type when legacy param is supplied.

# Submitter Checklist

# Release Notes

```release-note
If using prefix bucket, please pass // in LOGS_PATH, e.g.: LOGS_PATH=//logs
```

### DIFF
--- a/pkg/api/server/v1alpha2/plugin/plugin_logs.go
+++ b/pkg/api/server/v1alpha2/plugin/plugin_logs.go
@@ -39,9 +39,10 @@ import (
 )
 
 const (
-	lokiQueryPath   = "/loki/api/v1/query_range"
-	typePipelineRun = "tekton.dev/v1.PipelineRun"
-	typeTaskRun     = "tekton.dev/v1.TaskRun"
+	lokiQueryPath      = "/loki/api/v1/query_range"
+	typePipelineRun    = "tekton.dev/v1.PipelineRun"
+	typeTaskRun        = "tekton.dev/v1.TaskRun"
+	typeTaskRunV1Beta1 = "tekton.dev/v1beta1.TaskRun"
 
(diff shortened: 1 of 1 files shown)


khrm (commented):
/assign @enarha

khrm (commented):
/kind bug

pkg/api/server/v1alpha2/plugin/plugin_logs.go
```diff
@@ -304,10 +306,9 @@ func getBlobLogs(s *LogServer, writer io.Writer, parent string, rec *db.Record)
 		err := errors.New("pipelinerun not supported, please use taskrun")
 		s.logger.Error(err)
 		return err
-	case typeTaskRun:
-		logRec := &db.Record{}
-		if legacy {
-			logRec, err = getLogRecord(s.db, parent, rec.ResultID, rec.Name)
+	case typeTaskRunV1Beta1:
+		if legacy && rec.Type == typeTaskRunV1Beta1 {
```
enarha, line 310: Is `&& rec.Type == typeTaskRunV1Beta1` really required? Isn't that ensured by `case typeTaskRunV1Beta1`?

khrm: It's not required, I forgot to remove this.

khrm: Fixed this.

khrm:
@enarha Please review this again.

PR #1004, opened 2025-05-06
[CLI enhancement] Add `To be deprecated` message for the older commands

divyansh42:
# Changes
Follow-up PR towards the Results CLI enhancement TEP

[To be deprecated] will be shown only when the user is doing `--help`. This will not impact the usage of any of the existing commands. Future releases will mark these commands deprecated

# Submitter Checklist

# Release Notes

```release-note
Results, records and logs commands to be deprecated in the next release
```

### DIFF
--- a/pkg/cli/cmd/help.txt
+++ b/pkg/cli/cmd/help.txt
@@ -1,33 +1,21 @@
-Environment Variables:
-    TKN_RESULTS_SSL_ROOTS_FILE_PATH: Path to local SSL cert to use.
-    TKN_RESULTS_SSL_SERVER_NAME_OVERRIDE: SSL server name override (useful if using with a proxy such as kubectl port-forward).
+tkn-results CLI
 
-Config:
-    A config file may be stored in `~/.config/tkn/results.yaml` to configure the CLI client.
+tkn-results is a command-line interface (CLI) designed to interact with Tekton Results. This CLI provides tools to configure how you interact with the Tekton Results API server, query TaskRuns and PipelineRuns and their associated data.
 
-    Fields:
-    - address: Results API Server address
-    - service_account: When specified, the CLI will first fetch a bearer token
-                       for the specified ServiceAccount and attach that to Result API requests.
-        - namespace: ServiceAccount namespace
-        - name: ServiceAccount name
-    - token: Bearer token to use for API requests. Takes priority over service_account.
-    - ssl: SSL connection options
-        - roots_file_path: Path to a certificate to include in the cert pool. Useful for adding allowed self-signed certs.
-        - server_name_override: For testing only. Sets the grpc.ssl_target_name_override value for requests.
-    - portforward: enable auto portforwarding to tekton-results-api-service when address is set and portforward is true, tkn-results will portforward tekton-results-api-service automatically
-
-Example:
-
-    ```
-    address: results.dogfooding.tekton.dev:443
-    token: abcd1234
-    ssl:
-        roots_file_path: path/to/file
-        server_name_override: example.com
-    service_account:
-        namespace: default
-        name: result-reader
-    portforward: false
-    ```
-
+This CLI helps to:
+  config        Manage Tekton Results CLI configuration:
+                - set:  Configure the CLI to connect to a Tekton Results instance.
+                - view: Display the current CLI configuration.
+                - reset: Reset the CLI configuration to defaults.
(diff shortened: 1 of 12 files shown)


divyansh42:
@khrm @enarha should also add this in the release notes as well?

enarha:
> @khrm @enarha should also add this in the release notes as well?

IMO yes, definitely.

enarha (commented):
Given we want to deprecate this, we must provide some timeline of when we want to deprecate it. It could either be date or a version. I'm more in favor of the latter, which bring me to the next thing, perhaps we need to introduce versioning to the tool. It could be tight to the Tekton Results release version and we can also consider some number related to the offset from the commit from which the release was cut, if we want some sort of nightly version. For example if the latest version is `v0.14.0` and we are 7 commits above the release tag, the version could be ``v0.14.0-7`.
Which brings me to the next thing and it's releasing the tool in some fashion. All this should not be part of the current change, but something to consider and create issue(s) to be worked on. WDYT?

divyansh42:
> > @khrm @enarha should also add this in the release notes as well?
> 
> IMO yes, definitely.

Okay, I have added the release notes.

enarha:
> Given we want to deprecate this, we must provide some timeline of when we want to deprecate it. It could either be date or a version. I'm more in favor of the latter, which bring me to the next thing, perhaps we need to introduce versioning to the tool. It could be tight to the Tekton Results release version and we can also consider some number related to the offset from the commit from which the release was cut, if we want some sort of nightly version. For example if the latest version is `v0.14.0` and we are 7 commits above the release tag, the version could be ``v0.14.0-7`. Which brings me to the next thing and it's releasing the tool in some fashion. All this should not be part of the current change, but something to consider and create issue(s) to be worked on. WDYT?
> 
> Other than that, we can set some time/version for the actual deprecation. If we can decide about it now, then we can add it to the release notes. If not, we can add it to the release notes of a later change, once we know when the deprecation will actually happen.

oh, I see you mentioned next release, so is it going to be tight to the project release?

divyansh42:
> Given we want to deprecate this, we must provide some timeline of when we want to deprecate it. It could either be date or a version. I'm more in favor of the latter, which bring me to the next thing, perhaps we need to introduce versioning to the tool. It could be tight to the Tekton Results release version and we can also consider some number related to the offset from the commit from which the release was cut, if we want some sort of nightly version. For example if the latest version is `v0.14.0` and we are 7 commits above the release tag, the version could be ``v0.14.0-7`. Which brings me to the next thing and it's releasing the tool in some fashion. All this should not be part of the current change, but something to consider and create issue(s) to be worked on. WDYT?

Yes, that makes sense.
For the timeline, I mentioned that in the release notes; it will be deprecated in the next release. I can probably be more specific by specifying the exact version.

For the versioning: Yes, I have thought of that, but I haven't created a GH issue yet. There can be two ways, probably for that first one is about tying with the results release another one is having a separate versioning system and release independently. Let me know your thought on this?

enarha:
> > Given we want to deprecate this, we must provide some timeline of when we want to deprecate it. It could either be date or a version. I'm more in favor of the latter, which bring me to the next thing, perhaps we need to introduce versioning to the tool. It could be tight to the Tekton Results release version and we can also consider some number related to the offset from the commit from which the release was cut, if we want some sort of nightly version. For example if the latest version is `v0.14.0` and we are 7 commits above the release tag, the version could be ``v0.14.0-7`. Which brings me to the next thing and it's releasing the tool in some fashion. All this should not be part of the current change, but something to consider and create issue(s) to be worked on. WDYT?
> 
> Yes, that makes sense. For the timeline, I mentioned that in the release notes; it will be deprecated in the next release. I can probably be more specific by specifying the exact version.
> 
> For the versioning: Yes, I have thought of that, but I haven't created a GH issue yet. There can be two ways, probably for that first one is about tying with the results release another one is having a separate versioning system and release independently. Let me know your thought on this?

Lets have that discussion on the issue for versioning.
/lgtm

divyansh42:
FYI, Discussion thread for the binary release and versioning system:
https://github.com/tektoncd/results/issues/1007

khrm (commented):
/approve

khrm:
It makes sense to have a release note for this, but we don't release Cli as part of release.

divyansh42:
> It makes sense to have a release note for this, but we don't release Cli as part of release.

For now, I have added release notes for the users who are using this CLI

PR #1024, opened 2025-06-02
Fix usage of server related flags

divyansh42:
# Changes
Fix proper usage of server-related flags like `--host`, `--token` and `--api-path`
Earlier it was not getting used.

# Submitter Checklist

# Release Notes
```release-note
[CLI] Fix proper usage of server-related flags i.e. `--host, `--token`, `--api-path`, `--insecure-skip-tls-verify`
```

### DIFF
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -11,6 +11,7 @@ import (
 	"github.com/tektoncd/results/pkg/cli/client"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/spf13/cobra"
 	"github.com/tektoncd/results/pkg/cli/common"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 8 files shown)


divyansh42:
Will be including the test changes as a part of the test suite changes

pkg/cli/config/config.go
```diff
@@ -351,3 +352,56 @@ func getRawKubeConfigLoader(kubeconfigPath string) clientcmd.ClientConfig {
...
+	if host == "" || token == "" {
+		return nil, errors.New("--host and --token flag must be set if using direct connection flags")
+	}
+	rc := &rest.Config{
+		Host: host,
+	}
+	if p.APIPath() != "" {
+		rc.APIPath = p.APIPath()
+	}
+	if p.Token() != "" {
```
enarha, line 377: [nit] This check is not strictly necessary IMO as we check that in line 368. Then we can simply assign that value to the BearerToken field when we initialize the config on line 372.

pkg/cli/config/config.go
```diff
@@ -351,3 +352,56 @@ func getRawKubeConfigLoader(kubeconfigPath string) clientcmd.ClientConfig {
...
+		rc.BearerToken = token
+	}
+
+	rc.Insecure = p.SkipTLSVerify()
+	// Optionally set timeout (default 30s)
+	rc.Timeout = 30 * time.Second
```
enarha, line 383: I consider that default value low. I've witnessed some busy instances where it takes more than 30 second to get a response from the Results API service. Maybe 60 second. Even better if the user can modify that from the command line, could be done in a separate PR.

pkg/cli/config/config.go
```diff
@@ -351,3 +352,56 @@ func getRawKubeConfigLoader(kubeconfigPath string) clientcmd.ClientConfig {
...
+
+	rc.APIPath = path.Join(rc.APIPath, Path)
+
+	rc.GroupVersion = &schema.GroupVersion{
+		Group:   Group,
+		Version: Version,
+	}
+	u, pth, err := rest.DefaultServerUrlFor(rc)
```
enarha, line 391: note: that feels strange, the function returns 2 values (and an error), then we update the first value with the second one `u.Path = pth`. DefaultServerUrlFor needs some redesign IMO.

pkg/cli/config/config.go
```diff
@@ -351,3 +352,56 @@ func getRawKubeConfigLoader(kubeconfigPath string) clientcmd.ClientConfig {
 	// Return the clientcmd.ClientConfig (equivalent to ToRawKubeConfigLoader)
 	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
 }
+
+// ServerConnectionFlagsChanged returns true if any server connection flags are set.
+func ServerConnectionFlagsChanged(cmd *cobra.Command) bool {
```
enarha, line 357: ServerConnectionFlagsChanged and the following BuildDirectClientConfig are not used from outside the package. Do they have to be exported? Maybe if we add tests for them which we should probably do, then it makes sense.

pkg/cli/config/config.go
divyansh42: @enarha I am just using the `DefaultServerUrlFor` from the `client-go` library.

PR #1058, opened 2025-07-07
[CLI] Fetch routes only from specific namespaces in config set command

divyansh42:
# Changes

The `config set` command previously fetched routes from all namespaces, causing permission errors for non-admin users. This change modifies route detection to only check `openshift-pipelines` and `tekton-results` namespaces, with namespace existence validation before listing routes. The detection now filters by service name rather than labels, making it more robust and allowing non-admin users to use auto-detection with appropriate RBAC permissions. Auto-detection remains unsupported in Kubernetes environments, requiring manual configuration via flags with clear error messages guiding users to manual setup.
Support for the Kubernetes environment will be added in the future.

/kind bug

# Submitter Checklist

# Release Notes

```release-note
fix: [CLI] prevent permission errors for non-admin users during route auto-detection in config set command
```


docs/cli/tkn-results_config_set.md
```diff
@@ -7,74 +7,68 @@ Configure Tekton Results CLI settings
...
 
-  # Configure with custom API path and namespace (no prompts)
-  tkn-results config set --api-path=/api/v1 --namespace=my-namespace
+  # Manual configuration with custom settings
+  tkn-results config set --host=<api-server-url> --token=<token> --api-path=/api/v1 --insecure-skip-tls-verify
 
-  # Configure with custom kubeconfig and context
-  tkn-results config set --kubeconfig=/path/to/kubeconfig --context=my-cluster
+Automatic Detection (OpenShift only):
+- Detects routes in openshift-pipelines and tekton-results namespaces
```
aThorp96, line 32: The `tekton-results` namespace is not used by Openshift users generally; it is only in an Openshift Cluster when the admin has installed Tekton Results separately in a custom namespace.

divyansh42:
> I like the intention behind this change but I am not sure the implementation is as helpful as it seems. A couple things to note:
> 
> * This removes the non-openshift user's ability to detect the tekton-results endpoint. I don't think Openshift user's UX should come at the cost of non-openshift users' UX
> * The implementation checks the namespaces `openshift-pipelines` and `tekton-results`, however the `tekton-results` namespace is not created by Openshift Pipelines.
> 
> I would suggest refactoring this PR. First, we should keep or improve support for non-openshift users; at the very least we should keep the existing automated setup for non-openshift users. Better yet, iterate over known k8s and openshift-pipelines namespaces (`openshift-pipelines`, `tekton-results`, and `tekton-pipelines`, maybe others?) and check for either a Route (in the case of openshift) or a Service (for non-openshift)

Thanks @aThorp96 for the detailed feedback. I have re-designed the changes and pushed them to the PR. Please take a look.

pkg/cli/config/host.go
```diff
@@ -1,72 +1,123 @@
...
+	return "", fmt.Errorf("unable to extract cluster domain")
+}
+
+// isURLReachable checks if a URL is reachable with a simple HTTP request
+func isURLReachable(testURL string) bool {
+	client := &http.Client{
+		Timeout: 5 * time.Second,
+		// Allow insecure connections for testing
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
khrm: Yes. You can use nolint comment tag. It should fix that.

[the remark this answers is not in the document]
divyansh42: Done, thanks!

pkg/cli/config/platform.go
```diff
@@ -0,0 +1,73 @@
...
+		"security.openshift.io", // Security Context Constraints
+		"project.openshift.io",  // Projects
+		"user.openshift.io",     // Users and groups
+		"oauth.openshift.io",    // OAuth
+		"config.openshift.io",   // Cluster configuration
+	}
+
+	apiGroupList, err := discoveryClient.ServerGroups()
+	if err != nil {
+		return PlatformKubernetes // Default to Kubernetes if API discovery fails
```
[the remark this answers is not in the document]
divyansh42: I don't want to log anything as it will come to CLI output, and I want to do it silently.

khrm:
BTW, in future can you add test and doc generation in one commit. And implementation in another commit.

pkg/cli/cmd/config/set_test.go
```diff
@@ -1,143 +1,289 @@
...
 		},
 		{
-			name:       "nil params no flags",
-			params:     nil,
-			wantUse:    "set",
-			wantShort:  "Configure Tekton Results CLI settings",
-			flags:      map[string]string{},
-			wantPrompt: true,
-			wantErr:    false,
+			name:         "results-namespace only should prompt for host",
```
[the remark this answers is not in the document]
divyansh42: I think i have to remove those test cases as we have removed `results-namespace`
It got missed somehow.

khrm:
@enarha @aThorp96 Let's review and merge this asap.

docs/cli/tkn-results_config_set.md
```diff
@@ -1,39 +1,38 @@
...
-2. Prompt for any missing configuration values
-3. Save the configuration for future use
 
-Automatic Detection:
-- Cluster context and namespace
-- API server endpoint
-- Service account token (if available)
+Usage Modes:
+1. Interactive: Prompts for values with defaults where available
+   tkn-results config set
```
aThorp96, line 13: Is this supposed to be monospaced/code-format? The tab makes me think it is supposed to be, but it's rendering as paragraph text:
image

docs/man/man1/tkn-results-config-set.1
```diff
@@ -2,50 +2,46 @@
...
-  --token                   Authentication token
-  --api-path                API server path prefix
-  --insecure-skip-tls-verify Skip TLS certificate verification
-  --kubeconfig, -k          Path to kubeconfig file
-  --context, -c             Kubernetes context to use
-  --namespace, -n           Kubernetes namespace
+Route Requirements (OpenShift):
+- Route name: tekton-results-api-service
+- Namespace: openshift-pipelines
+- Expected URL format: https://-\&.apps.
```
aThorp96, line 41: ~I've never seen this format specification before. Can you clarify the format here with regex?~

I think since this line isn't a code block in the markdown the string `https://<route-name>-<namespace>.apps.<cluster-domain>` is being misconverted here in the manpage as `https://-\&.apps.`

pkg/cli/cmd/config/view.go
```diff
@@ -49,7 +48,22 @@ The configuration is displayed in YAML format.
...
+			}
+			if cmd.ErrOrStderr() != nil {
+				errOut = cmd.ErrOrStderr()
+			}
+
+			opts.IOStreams = &genericiooptions.IOStreams{
+				In:     os.Stdin,
+				Out:    out,
+				ErrOut: errOut,
+			}
```
aThorp96, line 65: Did you encounter an instance where `cmd.OutOrStdout()` or `cmd.ErrOrStderr()` returned `nil`? [Their implementation](https://github.com/spf13/cobra/blob/v1.9.1/command.go#L393) is such that if the cmd has no special out or err writer then they will return `os.Std{out,err}`. 

From what I can tell it should be safe and a bit simpler to do:
```suggestion
			opts.IOStreams = &genericiooptions.IOStreams{
				In:     cmd.InOrStdin(),
				Out:    cmd.OutOrStdout(),
				ErrOut: cmd.ErrOrStderr() ,
			}
```

aThorp96 (commented):
Haven't been able to go through the whole thing yet but there were a couple notes on the docs formatting and pre-run

pkg/cli/config/platform.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	availableGroups := make(map[string]bool)
+	for _, group := range apiGroupList.Groups {
+		availableGroups[group.Name] = true
+	}
+
+	// Check if any OpenShift API groups are present
+	for _, osGroup := range openShiftAPIGroups {
+		if availableGroups[osGroup] {
```
pratap0007, line 54:
```suggestion
		if _, found := availableGroups[osGroup]; found {
```

pkg/cli/config/config.go
```diff
@@ -197,9 +196,6 @@ func (c *config) Set(prompt bool, p common.Params) error {
 	// get data from prompt in enabled
 	if prompt {
 		host := c.Host()
-		if err, ok := host.(error); ok {
```
pratap0007, line 200: why did we remove this condition?

pkg/cli/cmd/config/config_test.go
```diff
@@ -3,221 +3,87 @@ package config
...
-			assert.Equal(t, tt.expectedUse, cmd.Use)
-			assert.Equal(t, tt.expectedShort, cmd.Short)
-
-			// Verify flags
-			for _, flagName := range tt.expectedFlags {
-				flag := cmd.PersistentFlags().Lookup(flagName)
-				assert.NotNil(t, flag, "Expected flag %s to exist", flagName)
+			// Verify basic command properties
+			if cmd.Use != "config" {
+				t.Errorf("Expected Use to be %q, got %q", "config", cmd.Use)
```
pratap0007, line 34:
```suggestion
				t.Errorf("expected Use to be %q, got %q", "config", cmd.Use)
```

pkg/cli/cmd/config/config_test.go
```diff
@@ -3,221 +3,87 @@ package config
...
+			}
+			if cmd.Short != "Manage Tekton Results CLI configuration" {
+				t.Errorf("Expected Short to be %q, got %q", "Manage Tekton Results CLI configuration", cmd.Short)
```
pratap0007, line 37:
```suggestion
				t.Errorf("expected Short to be %q, got %q", "Manage Tekton Results CLI configuration", cmd.Short)
```

pkg/cli/cmd/config/config_test.go
```diff
@@ -3,221 +3,87 @@ package config
...
 			}
 
-			// Verify subcommands
+			// Verify specific subcommands exist
 			subcmds := cmd.Commands()
-			assert.Len(t, subcmds, len(tt.expectedSubcmds))
-			for _, expectedSubcmd := range tt.expectedSubcmds {
+			expectedSubcmds := []string{"set", "reset", "view"}
+			if len(subcmds) != len(expectedSubcmds) {
+				t.Errorf("Expected %d subcommands, got %d", len(expectedSubcmds), len(subcmds))
```
pratap0007, line 44:
```suggestion
				t.Errorf("expected %d subcommands, got %d", len(expectedSubcmds), len(subcmds))
```

pkg/cli/cmd/config/config_test.go
```diff
@@ -3,221 +3,87 @@ package config
...
-			// Command should always be created, even with nil params
-			require.NotNil(t, cmd, "Command should not be nil even for nil params")
-			require.NotNil(t, cmd.PersistentPreRunE, "PersistentPreRunE should not be nil")
-
-			// Parse flags before running PersistentPreRunE
-			cmd.SetArgs([]string{})
-			if err := cmd.ParseFlags([]string{}); err != nil {
-				t.Fatalf("Failed to parse flags: %v", err)
+				if !found {
+					t.Errorf("Expected subcommand %q to exist", expectedCmd)
```
pratap0007, line 56:
```suggestion
					t.Errorf("expected subcommand %q to exist", expectedCmd)
```

pkg/cli/cmd/config/config_test.go
```diff
@@ -3,221 +3,87 @@ package config
...
 
-			// Only test the PersistentPreRunE function if we have a command
-			err := cmd.PersistentPreRunE(cmd, []string{})
-			if tt.wantErr {
-				assert.Error(t, err, "Expected error for nil params")
-			} else {
-				assert.NoError(t, err, "Expected no error for valid params")
+			// Verify PersistentPreRunE is set
+			if cmd.PersistentPreRunE == nil {
+				t.Error("Expected PersistentPreRunE to be set")
```
pratap0007, line 62:
```suggestion
				t.Error("expected PersistentPreRunE to be set")
```

pkg/cli/cmd/config/config_test.go
```diff
@@ -3,221 +3,87 @@ package config
...
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			flag := cmd.PersistentFlags().Lookup(tt.flagName)
-			require.NotNil(t, flag)
-			assert.Equal(t, tt.shorthand, flag.Shorthand)
-		})
+	// Test that it doesn't error with valid setup
+	err = cmd.PersistentPreRunE(cmd, []string{})
+	if err != nil {
+		t.Errorf("Expected no error for valid params, got: %v", err)
```
pratap0007, line 87:
```suggestion
		t.Errorf("expected no error for valid params, got: %v", err)
```

pkg/cli/cmd/config/reset_test.go
```diff
@@ -4,60 +4,116 @@ import (
...
 			}
 
-			// Verify command properties
-			if cmd.Use != tt.wantUse {
-				t.Errorf("Expected Use to be %q, got %q", tt.wantUse, cmd.Use)
-			}
-			if cmd.Short != tt.wantShort {
-				t.Errorf("Expected Short to be %q, got %q", tt.wantShort, cmd.Short)
+			if cmd.Use != "reset" {
+				t.Errorf("Expected command use to be 'reset', got %q", cmd.Use)
```
pratap0007, line 38:
```suggestion
				t.Errorf("expected command use to be 'reset', got %q", cmd.Use)
```

pkg/cli/cmd/config/reset_test.go
```diff
@@ -4,60 +4,116 @@ import (
...
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			// Create the command
 			cmd := resetCommand(tt.params)
+
 			if cmd == nil {
-				t.Fatal("Expected command to be created")
+				t.Error("Expected command to be created")
```
pratap0007, line 33:
```suggestion
				t.Error("expected command to be created")
```

pkg/cli/cmd/config/set_test.go
```diff
@@ -1,143 +1,245 @@
...
+			allArgs := append([]string{"--kubeconfig=" + kubeconfigPath}, tt.args...)
+			cmd.SetArgs(allArgs)
+
+			// Execute the command
+			err := cmd.Execute()
+
+			// Check error expectations
+			if tt.expectError {
+				if err == nil {
+					t.Errorf("%s: Expected error but got none", tt.description)
```
pratap0007, line 83:
```suggestion
					t.Errorf("%s: expected error but got none", tt.description)
```

pkg/cli/cmd/config/set_test.go
```diff
@@ -1,143 +1,245 @@
...
+				}
+				return
 			}
 
-			// Verify command properties
-			if cmd.Use != tt.wantUse {
-				t.Errorf("Expected Use to be %q, got %q", tt.wantUse, cmd.Use)
+			// Should not have error for success cases
+			if err != nil {
+				t.Errorf("%s: Unexpected error: %v", tt.description, err)
```
pratap0007, line 90:
```suggestion
				t.Errorf("%s: unexpected error: %v", tt.description, err)
```

pkg/cli/cmd/config/set_test.go
```diff
@@ -1,143 +1,245 @@
...
+
+			if tt.expectPrompt && !containsPrompt {
+				t.Errorf("%s: Expected prompting but found no prompt messages in output", tt.description)
 			}
 
-			// Initialize the flags by parsing empty args
-			if err := cmd.ParseFlags([]string{}); err != nil {
-				t.Fatalf("Failed to parse flags: %v", err)
+			if !tt.expectPrompt && containsPrompt {
+				t.Errorf("%s: Expected no prompting but found prompt messages in output", tt.description)
```
pratap0007, line 104:
```suggestion
				t.Errorf("%s: expected no prompting but found prompt messages in output", tt.description)
```

pkg/cli/cmd/config/view_test.go
```diff
@@ -1,63 +1,164 @@
...
 
-			// Verify command properties
-			if cmd.Use != tt.wantUse {
-				t.Errorf("Expected Use to be %q, got %q", tt.wantUse, cmd.Use)
+			// Now test view command and capture output
+			cmd := Command(params)
+			viewArgs := []string{"--kubeconfig=" + kubeconfigPath, "view"}
+			output, err := testutil.ExecuteCommand(cmd, viewArgs...)
+			if err != nil {
+				t.Errorf("%s: Unexpected error: %v", tt.description, err)
```
pratap0007, line 108:
```suggestion
				t.Errorf("%s: unexpected error: %v", tt.description, err)
```

pkg/cli/cmd/config/view_test.go
```diff
@@ -1,63 +1,164 @@
...
-				t.Errorf("Expected Short to be %q, got %q", tt.wantShort, cmd.Short)
+
+			// Verify output contains expected values
+			t.Logf("View command output:\n%s", output)
+
+			// Check for expected host
+			if tt.expectHost != "" {
+				expectedHostLine := "host: " + tt.expectHost
+				if !strings.Contains(output, expectedHostLine) {
+					t.Errorf("%s: Expected output to contain %q, but it was missing.\nFull output:\n%s",
```
pratap0007, line 119:
```suggestion
					t.Errorf("%s: expected output to contain %q, but it was missing.\nFull output:\n%s",
```

pkg/cli/cmd/config/view_test.go
```diff
@@ -1,63 +1,164 @@
...
+			} else if !strings.Contains(output, `token: ""`) {
+				t.Errorf("%s: Expected output to contain empty token, but it was missing.\nFull output:\n%s",
+					tt.description, output)
+			}
+
+			// Check for expected API path (only if specified)
+			if tt.expectAPIPath != "" {
+				expectedAPIPathLine := "api-path: " + tt.expectAPIPath
+				if !strings.Contains(output, expectedAPIPathLine) {
+					t.Errorf("%s: Expected output to contain %q, but it was missing.\nFull output:\n%s",
```
pratap0007, line 143:
```suggestion
					t.Errorf("%s: expected output to contain %q, but it was missing.\nFull output:\n%s",
```

pratap0007:
LGTM

docs/man/man1/tkn-results-config-set.1
divyansh42: Nice catch! Fixed.

pratap0007:
LGTM

pkg/cli/config/host.go
```diff
@@ -1,72 +1,130 @@
...
+		return false
+	}
+	defer func() {
+		_ = resp.Body.Close() // Ignore close error for connectivity test
+	}()
+
+	// Consider any HTTP response (even errors like 401 Unauthorized, 403 Forbidden) as "reachable"
+	// because these indicate the service is present and responding, but may require authentication.
+	// This allows us to detect connectivity to the service even if credentials are missing or invalid.
+	return resp.StatusCode < 500 // 2xx, 3xx, 4xx (including auth errors) are all considered reachable
```
aThorp96, line 129: Nonblocking suggestion:
If we're just testing that the given hostname is reachable, we don't even need the http layer by using [net.DialTimeout](https://godocs.io/net#DialTimeout) to test that we can dial the url:
```suggestion
	conn, err := net.DialTimeout("tcp", testURL, t * time.Second)
	if conn != nil {
		conn.Close()
	}
	return err == nil
```

aThorp96, line 129: Is a 500, 502, or 504 considered unreachable here? If I understand correctly that means the host _exists_ and we can _reach it_ but the Tekton Results API may be down. I think still tells us the URL is "reachable" in the sense we need here, but maybe that isn't suitable for other reasons?

pkg/cli/config/platform.go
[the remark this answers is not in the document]
aThorp96: I think this is a valid point actually. What does an error here actually indicate? To me it stands to reason that we should be returning `PlatformUnknown` in this case

pkg/cli/config/platform.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, group := range apiGroupList.Groups {
+		availableGroups[group.Name] = true
+	}
+
+	// Check if any OpenShift API groups are present
+	for _, osGroup := range openShiftAPIGroups {
+		if _, found := availableGroups[osGroup]; found {
+			return PlatformOpenShift
+		}
+	}
```
aThorp96, line 57: If `openShiftAPIGroups` is not defined as a list but as a `map[string]bool` (God I wish Golang had proper Set data structures), then we can short this loop without needing to iterate over every API group in the cluster:
```go
	openShiftAPIGroups := map[string]bool{
		"route.openshift.io": true,    // Routes (core OpenShift feature)
		"image.openshift.io": true,    // Image streams
		"apps.openshift.io": true,     // DeploymentConfigs
		"security.openshift.io": true, // Security Context Constraints
		"project.openshift.io": true,  // Projects
		"user.openshift.io": true,     // Users and groups
		"oauth.openshift.io": true,    // OAuth
		"config.openshift.io": true,   // Cluster configuration
	}
	...

	for _, group := range apiGroupList.Groups {
		if _, ok := openShiftAPIGroups[group]; ok {
			return PlatformOpenShift
	}

	return PlatformKubernetes
```

pkg/cli/testutils/kubeconfig.go
```diff
@@ -0,0 +1,139 @@
...
+
+	cleanup := func() {
+		_ = os.Remove(tmpFile.Name()) // Ignore error for test cleanup
+	}
+
+	return tmpFile.Name(), cleanup
+}
+
+// CreateTestKubeconfigForExecution creates a temporary kubeconfig file for functional testing
+func CreateTestKubeconfigForExecution(t *testing.T) (string, func()) {
```
aThorp96, line 56: `CreateTestKubeconfigForExecution` and `CreateTestKubeconfig` appear functionally identical. Can we consolidate the two?

pkg/cli/testutils/kubeconfig.go
```diff
@@ -0,0 +1,139 @@
...
+	t.Helper()
+
+	dir, err := os.MkdirTemp("", "kubeconfig-test")
```
aThorp96, line 59: Please use `t.TempDir()` instead of `os.MkdirTemp()` . This has the added benefit that you'll no longer need the cleanup func: it automatically removes the directory once the current `Testing.T` is complete. See https://godocs.io/testing#T.TempDir

aThorp96 (commented):
Looks generally good. I have a few small suggestions

pkg/cli/config/host.go
```diff
@@ -1,72 +1,130 @@
...
 	}
 
-	routeV1Client, err := routev1.NewForConfig(c)
+	platform := DetectPlatform(c)
+
+	switch platform {
+	case PlatformOpenShift:
+		return tryConnectToRoute(c)
+	case PlatformKubernetes:
+		return "", fmt.Errorf("kubernetes ingress not supported")
```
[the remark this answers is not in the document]
divyansh42: We have discussed and tried to support non-OpenShift env as well, but it doesn't seem to be very direct. We are looking for a specific pattern for the host URL and with the OpenShift env, we are only supporting the pattern which is available by default. i.e. `https://<route-name>-<namespace>.apps.<cluster-domain>`. If any setup has a host different from this format, then they will have to manually enter the host. 

The thing we have improved for all the platforms is that earlier there was no option to add the host manually if you are not using the flags, but now you can modify the host that is shown in the prompt or if not shown you can enter the host.

We were not able to do auto detection in the case of k8s, because we were not aware of any set pattern for the ingress and we do not want to introduce the complexities where we list all the ingress in the namespaces.

pkg/cli/config/platform.go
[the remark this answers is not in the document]
divyansh42: Do you think we should fail the command if we get PlatformUnknown?

pkg/cli/testutils/kubeconfig.go
divyansh42: Fixed.

divyansh42:
@aThorp96 thanks for the suggestions. I have tried to incorporate most of them. I have added a comment where I had some doubt or needed clarification.
Can you please take a look?

PR #1078, opened 2025-08-04
Refactor annotations package and switch to Server-Side Apply

enarha:
Removed business logic from the annotation package. Now it only provides functionality to manage annotations, but no logic when to set them. Consolidated check (if to apply), generation and actual object update. Before that change, the caller had to check if the patch is needed, call the generation of the patch and apply the patch. Now it's a single call to the Patch function.
Switched to Server-Side Apply patches from Merge patches, which provide a more robust way to update k8s objects.
Switched clients.go to a separate package to avoid cyclic imports.

# Changes

/kind design
# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/watcher/reconciler/annotation/annotation.go
+++ b/pkg/watcher/reconciler/annotation/annotation.go
@@ -15,42 +15,55 @@
 package annotation
 
 import (
+	"context"
 	"encoding/json"
+	"fmt"
+	"strings"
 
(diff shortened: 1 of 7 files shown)


vdemeester (approved):

khrm (commented):
/lgtm

khrm:
/kind feature

I feel it should go under feature in release note.

PR #1187, opened 2026-01-29
Add explicit TLS configuration support

enarha:
This is to help the centralized TLS management support. It allows the Tekton Operator to propagate TLS configuration to all operands in the format of env variables. This change also allows the administrator to set TLS configuration through the API configuration. In either case the all-or-nothing approach is followed, all three configuration options are always taken from the same source without mixing them.

# Changes

/kind feature
# Submitter Checklist

# Release Notes

```release-note
Added support for flexible TLS configuration through environment variables. This allows TLS settings to be managed externally (e.g., via Kubernetes ConfigMaps or the Tekton Operator) without requiring code changes. The following three configuration options were added:
- TLS_MIN_VERSION - Minimum TLS version
- TLS_CIPHER_SUITES - Comma-separated cipher suites (IANA names or numeric IDs)
- TLS_CURVE_PREFERENCES - Comma-separated elliptic curves for key exchange

See [TLS Configuration](./cmd/api/README.md#tls-configuration) for more details.
```


cmd/api/main.go
```diff
@@ -91,14 +93,51 @@ func main() {
...
+		tlsConf = tlsconfig.LoadFromEnv(os.Getenv)
+		log.Debug("TLS configuration loaded from environment variables")
+	} else {
+		tlsConf = tlsconfig.New(serverConfig.TLS_MIN_VERSION, serverConfig.TLS_CIPHER_SUITES, serverConfig.TLS_CURVE_PREFERENCES)
+		log.Debug("TLS configuration loaded from config file")
+	}
+	serverTLSConfig, tlsConfErr := tlsConf.ToTLSConfig()
+	if tlsConfErr != nil {
+		log.Errorf("Error parsing TLS configuration: %v", tlsConfErr)
+		log.Warn("Using default TLS configuration")
```
khrm: Shouldn't we panic in case of wrong config? It means the user has configured something wrong, and the default setup might have some security assumptions.

config/base/env/config
```diff
@@ -18,6 +18,9 @@ SERVER_PORT=8080
 PROMETHEUS_PORT=9090
 PROMETHEUS_HISTOGRAM=false
 TLS_PATH=/etc/tls
+TLS_MIN_VERSION=
```
khrm: Should we put some default values here which are mentioned in doc?

enarha: The values from the doc are not the default values, but just an example moving "higher" from the default values. The default values (when the configuration options are omitted are the default values from Go, e.g. as of Go 1.18 the TLS min version is 1.2). I had to explicitly set that version in the code because CodeQL is complaining when it's not explicitly set. Other than that I think it makes sense to stick to the Go values and I do not like to hardcode them in the config as they might change with the newer versions of Go.

khrm (commented):
/lgtm

Issue #1204, opened 2026-02-10
Add Configurable Log Output Formatting Including Timestamps


0x49b:
### Feature request

We would like the ability to configure the line_format of the query that returns log results. Currently, the result string includes only the message, or when configured the container key (LOGGING_PLUGIN_CONTAINER_KEY). We also need the timestamp to be included, which is available in the raw log under timestamp.

This could be implemented more broadly so that, in addition to the timestamp, other labels could also be used in the formatted output for additional use cases.

If the configuration Option is not set, the current default is used.

### Use case

We provide a custom application for builds using Tekton. We are currently implementing the Loki stack to store results in Loki instead of Postgres. However, when retrieving results from Loki, the timestamp is missing, which prevents us from ordering the results correctly for display to users.

thomascube:
While the PR is merged, the issue is still open. Is this feature part of the 0.19.0 release? If so, it's not mentioned in the release notes.

enarha:
> While the PR is merged, the issue is still open. Is this feature part of the 0.19.0 release? If so, it's not mentioned in the release notes.

Yes, it's part of the release. I see the commit in the release branch: https://github.com/tektoncd/results/tree/release-v0.19.x
It is not included in the release notes because the kind label was not properly set. You can see how it should be here: https://github.com/tektoncd/results/pull/1320 (see the kind/feature label) compared to what we have here https://github.com/tektoncd/results/pull/1210. The `/kind feature` text is there in the PR description, but it was not picked up by the bot, so the label used for the release text generation is missing. Not sure how that happened, probably some timing issue, but I do not have enough data to verify. Anyway, I believe I can add that manually. Thanks for reporting it.

enarha:
> > While the PR is merged, the issue is still open. Is this feature part of the 0.19.0 release? If so, it's not mentioned in the release notes.
> 
> Yes, it's part of the release. I see the commit in the release branch: https://github.com/tektoncd/results/tree/release-v0.19.x It is not included in the release notes because the kind label was not properly set. You can see how it should be here: [#1320](https://github.com/tektoncd/results/pull/1320) (see the kind/feature label) compared to what we have here [#1210](https://github.com/tektoncd/results/pull/1210). The `/kind feature` text is there in the PR description, but it was not picked up by the bot, so the label used for the release text generation is missing. Not sure how that happened, probably some timing issue, but I do not have enough data to verify. Anyway, I believe I can add that manually. Thanks for reporting it.

Added to the release notes.

PR #1210, opened 2026-02-11
Added config options to define json fields and line_format for loki logs

0x49b:
# Changes

Added configuration variables to define a json map to extract fields and to define the line_format of Loki logs. By passing a string with json in `LOGGING_PLUGIN_JSON_MAP` like `{"timestamp": "[\"@timestamp\"]"}` the json field timestamp gets extracted into timestamp what then can be used to define a custom line_format in `LOGGING_PLUGIN_LINE_FORMAT` like 
`"{{.timestamp}}: {{.message}}"`

Mentioned in #1204 

/kind feature

# Submitter Checklist

# Release Notes

```release-note
Make the Loki LogQL more configurable by letting the user decide a json map for fields to be extracted and also define the following line_format (#1204) 
```

### DIFF
--- a/docs/logging-support.md
+++ b/docs/logging-support.md
@@ -58,3 +58,7 @@ These are the common configuration options for all third party logging APIs.
 - `LOGGING_PLUGIN_QUERY_LIMIT`: Sets the query limit for Third Party Logging API if logging backend has a limit on number of log lines returned.
 - `LOGGING_PLUGIN_QUERY_PARAMS`: Sets the query params for Third Party Logging API, these can be direction/sort order.Specify them in this format: "foo=bar&direction=backward"
 - `LOGGING_PLUGIN_MULTIPART_REGEX`: Sets a Regex for matching parts of the same log. Some log backends (e.g S3) store objects immutably, once stored, you can't append. For long running TaskRun steps, it's not effective to keep such log in memory until the step completes. Instead one can store the log in multiple parts with a name suffix (e.g `-1743932245` seconds since the Epoch) and set a regex to match the parts of the same log (e.g `-\d{10}$`). (optional)
+
+## Loki-specific Configuration
+- `LOGGING_PLUGIN_JSON_MAP`: Define a map for the fields in Loki logs to be extracted like `{"timestamp": "@timestamp"}` to extract the timestamp as example.
+- `LOGGING_PLUGIN_LINE_FORMAT`: Define the format of the log line returned when using Loki. To use the Timestamp from the `LOGGING_PLUGIN_JSON_MAP` and the actual message, define `"{{.timestamp}}: {{.message}}"`. Only fields extracted by `LOGGING_PLUGIN_JSON_MAP` can be used here.
--- a/pkg/api/server/v1alpha2/plugin/plugin_logs.go
+++ b/pkg/api/server/v1alpha2/plugin/plugin_logs.go
@@ -185,10 +185,53 @@ func getLokiLogs(s *LogServer, writer io.Writer, parent string, rec *db.Record) 
 	for k, v := range s.queryParams {
 		parameters.Add(k, v)
 	}
-	query := `{ ` + s.staticLabels + s.config.LOGGING_PLUGIN_NAMESPACE_KEY + `="` + parent + `" }|json uid="` + uidKey + `", message="message" |uid="` + rec.Name + `"| line_format "{{.message}}"`
-	if s.config.LOGGING_PLUGIN_CONTAINER_KEY != "" {
-		query = `{ ` + s.staticLabels + s.config.LOGGING_PLUGIN_NAMESPACE_KEY + `="` + parent + `" }|json uid="` + uidKey + `", container="` + s.config.LOGGING_PLUGIN_CONTAINER_KEY + `", message="message" |uid="` + rec.Name + `"| line_format "container-{{.container}}: message={{.message}}"`
+
+	baseSelector := fmt.Sprintf(`{ %s%s="%s" }`, s.staticLabels, s.config.LOGGING_PLUGIN_NAMESPACE_KEY, parent)
+	uidFilter := fmt.Sprintf(`|uid="%s"`, rec.Name)
+	jsonFields := map[string]string{
+		"uid":     uidKey,
+		"message": "message",
 	}
+
+	if s.config.LOGGING_PLUGIN_CONTAINER_KEY != "" {
+		jsonFields["container"] = s.config.LOGGING_PLUGIN_CONTAINER_KEY
+	}
+
+	if s.config.LOGGING_PLUGIN_JSON_MAP != "" {
+		var userMap map[string]string
+		if err := json.Unmarshal([]byte(s.config.LOGGING_PLUGIN_JSON_MAP), &userMap); err != nil {
+			s.logger.Errorf("Error parsing LOGGING_PLUGIN_JSON_MAP: %v", err)
+		} else {
+			for k, v := range userMap {
+				jsonFields[k] = v
+			}
+		}
+	}
+	var keys []string
+	for k := range jsonFields {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var jsonArgs []string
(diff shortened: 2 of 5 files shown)


pkg/api/server/v1alpha2/plugin/plugin_logs.go
```diff
@@ -185,10 +185,53 @@ func getLokiLogs(s *LogServer, writer io.Writer, parent string, rec *db.Record)
...
+			for k, v := range userMap {
+				jsonFields[k] = v
+			}
+		}
 	}
+	var keys []string
+	for k := range jsonFields {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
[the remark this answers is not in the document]
0x49b: For the functionality itself it is not necessary to sort the json properties. But Loki calculates a hash for the query which is then used to do caching. In Go the iteration order of a range over a map is not specified and can be different each time. So if the labels are not sorted Loki caches several entries for same query. When running a sort, we can avoid that having a cleaner cache and better response time. The "problem" is not on line 211/212 but on 205 where it iterates over a `userMap map[string]string`. 

https://grafana.com/docs/loki/latest/operations/caching/#results-cache
https://go.dev/ref/spec#For_range (at 3.)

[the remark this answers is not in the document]
thomascube: That makes perfect sense, thanks for the explanation!

aThorp96 (commented):
This seems like a good change to me. Thanks for the contribution! 

A couple asks:
- This is configured using a vendor-agnostic variable `LOGGING_PLUGIN_LINE_FORMAT`. It makes sense to me that this feature should be implemented for the other logging plugin formats supported by `LOGS_TYPE` or at least issues created to port support to the other log types (I believe the full list of log types are Loki, GCS, S3, "blob" (same as S3?), and Splunk). 
- Please also add unit tests and documentation (docs/logging-supprot.md) for the new configuration options. It's worth noting in that documentation that `LOGGING_PLUGIN_JSON_MAP` has no effect on the logging output unless used in conjunction with `LOGGING_PLUGIN_LINE_FORMAT`.

Along a similar lines, since I am not very familiar with Loki: since there's the `json key=value` component to the query, does `LOGGING_PLUGIN_LINE_FORMAT` work when used without `LOGGING_PLUGIN_JSON_MAP`, or do they have to be used together in all cases?

docs/logging-support.md
```diff
@@ -58,3 +58,7 @@ These are the common configuration options for all third party logging APIs.
 - `LOGGING_PLUGIN_QUERY_LIMIT`: Sets the query limit for Third Party Logging API if logging backend has a limit on number of log lines returned.
 - `LOGGING_PLUGIN_QUERY_PARAMS`: Sets the query params for Third Party Logging API, these can be direction/sort order.Specify them in this format: "foo=bar&direction=backward"
 - `LOGGING_PLUGIN_MULTIPART_REGEX`: Sets a Regex for matching parts of the same log. Some log backends (e.g S3) store objects immutably, once stored, you can't append. For long running TaskRun steps, it's not effective to keep such log in memory until the step completes. Instead one can store the log in multiple parts with a name suffix (e.g `-1743932245` seconds since the Epoch) and set a regex to match the parts of the same log (e.g `-\d{10}$`). (optional)
+
+## Loki-specific Configuration
+- `LOGGING_PLUGIN_JSON_MAP`: Define a map for the fields in Loki logs to be extracted like `{"timestamp": "[\"@timestamp\"]"}` to extract the timestamp as example.
```
khrm: JSON. Unmarshal into map[string]string so values are plain strings. Shouldn't this be {"timestamp": "@timestamp"}

0x49b: You're right, i guess a copy paste fail

khrm:
@0x49b Can you please sign easy cla? And squash the commits?

0x49b:
EasyCLa for me (0x49b) is done, but the one for @bhanuprasad14 is not. Not a Colleague of mine... Squash I will do

0x49b:
@khrm Did the squash, this also solved the issue with the not signed CLA

khrm (commented):
/approve

@enarha @aThorp96 Let's merge this.

thomascube:
@khrm What's the status of this PR? Is there still anything missing? We're eagerly waiting for this feature to be released. Thanks!

kyubisation:
@khrm Hello there 👋 
Another week has passed without an update. Is there still a problem? We would really love to use this.

enarha:
Apologies for the delay. I'm having a look and if nothing critical I'll merge and include in the upcoming release.

docs/logging-support.md
```diff
@@ -58,3 +58,7 @@ These are the common configuration options for all third party logging APIs.
...
 - `LOGGING_PLUGIN_QUERY_PARAMS`: Sets the query params for Third Party Logging API, these can be direction/sort order.Specify them in this format: "foo=bar&direction=backward"
 - `LOGGING_PLUGIN_MULTIPART_REGEX`: Sets a Regex for matching parts of the same log. Some log backends (e.g S3) store objects immutably, once stored, you can't append. For long running TaskRun steps, it's not effective to keep such log in memory until the step completes. Instead one can store the log in multiple parts with a name suffix (e.g `-1743932245` seconds since the Epoch) and set a regex to match the parts of the same log (e.g `-\d{10}$`). (optional)
+
+## Loki-specific Configuration
+- `LOGGING_PLUGIN_JSON_MAP`: Define a map for the fields in Loki logs to be extracted like `{"timestamp": "@timestamp"}` to extract the timestamp as example.
+- `LOGGING_PLUGIN_LINE_FORMAT`: Define the format of the log line returned when using Loki. To use the Timestamp fron the `LOGGING_PLUGIN_JSON_MAP` and the actual message, define `"{{.timestamp}}: {{.message}}"`. Only fields extracted by `LOGGING_PLUGIN_JSON_MAP` can be used here.
```
enarha: typo: *from*

kyubisation:
Thank you very much 🙇

pkg/api/server/v1alpha2/plugin/plugin_logs.go
```diff
@@ -185,10 +185,53 @@ func getLokiLogs(s *LogServer, writer io.Writer, parent string, rec *db.Record)
...
+
+	var jsonArgs []string
+	for _, k := range keys {
+		escapedVal := strings.ReplaceAll(jsonFields[k], `"`, `\"`)
+		jsonArgs = append(jsonArgs, fmt.Sprintf(`%s="%s"`, k, escapedVal))
```
enarha, line 219: The values are escaped, but not the keys. Theoretically one could set `LOGGING_PLUGIN_JSON_MAP='{"my field":"@timestamp"}'` and that will result `| json my field="@timestamp"`. Isn't it something to cover as well?

docs/logging-support.md
0x49b: Typo corrected

pkg/api/server/v1alpha2/plugin/plugin_logs.go
0x49b: Yes, the user can configure a key with a space, but Loki does not produce any label keys with spaces. Regarding the docs 

> Extracted label keys are automatically sanitized by all parsers, to follow Prometheus metric name convention.(They can only contain ASCII letters and digits, as well as underscores and colons. They cannot start with a digit.)

Source: https://grafana.com/docs/loki/latest/query/log_queries/#parser-expression and https://prometheus.io/docs/practices/naming/

Also it is not allowed to define custom labels with any special characters (like /) or spaces https://grafana.com/docs/loki/latest/get-started/labels/#:~:text=It%20may%20contain%20ASCII%20letters,should%20be%20written%20as%20app_kubernetes_io_name%20.

enarha: OK. Thanks. Please squash the 2 commits and I'll merge.

0x49b:
@enarha squashed those commits

PR #1285, opened 2026-04-01
fix(watcher): populate RecordSummary StartTime/EndTime from Status fields

SergK:
`RecordSummary.StartTime` is always NULL and `EndTime` is NULL for
failed/timed-out/cancelled runs.

The watcher derived timestamps from Knative conditions via `getTimestamp()`.
This is broken for two reasons:

1. **StartTime** was read from `ConditionReady`. Tekton PipelineRun and TaskRun
   are batch resources (`NewBatchConditionSet`) whose happy condition is
   `ConditionSucceeded`. `ConditionReady` is for living resources like Knative
   Services and is never set on batch resources, so the lookup always returned
   nil.

2. **EndTime** was read from `ConditionSucceeded`, but `getTimestamp()` returned
   nil when the condition `IsFalse()`. For failed, timed-out, and cancelled runs
   `ConditionSucceeded` is False, so EndTime was nil even though the run had
   completed.

## Fix

Read `Status.StartTime` and `Status.CompletionTime` directly from the
PipelineRun/TaskRun object via type switch. These fields are set by the Tekton
controller for all terminal states. This follows the existing pattern used by
`getCompletionTime()` in the dynamic reconciler.

- Add `getStartTime()` and `getEndTime()` helpers
- Update `ensureResult()` to use the new helpers
- Remove the now-unused `getTimestamp()` function
- Add unit and integration tests covering both run types in success and failure
  states

/kind bug

# Submitter Checklist

# Release Notes

```release-note
Fix RecordSummary StartTime always being NULL and EndTime being NULL for
failed/timed-out/cancelled runs. Timestamps are now read directly from the
run's Status.StartTime and Status.CompletionTime fields instead of from
Knative conditions.
```

### DIFF
--- a/pkg/watcher/results/results.go
+++ b/pkg/watcher/results/results.go
@@ -23,6 +23,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/google/go-cmp/cmp"
+	pipelinev1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	"github.com/tektoncd/results/pkg/api/server/v1alpha2/record"
 	"github.com/tektoncd/results/pkg/api/server/v1alpha2/result"
 	"github.com/tektoncd/results/pkg/watcher/convert"
@@ -124,8 +125,8 @@ func (c *Client) ensureResult(ctx context.Context, o Object, opts ...grpc.CallOp
 			Record:    recName,
 			Type:      convert.TypeName(o),
 			Status:    convert.Status(o.GetStatusCondition()),
-			StartTime: getTimestamp(o.GetStatusCondition().GetCondition(apis.ConditionReady)),
-			EndTime:   getTimestamp(o.GetStatusCondition().GetCondition(apis.ConditionSucceeded)),
+			StartTime: getStartTime(o),
+			EndTime:   getEndTime(o),
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


khrm (commented):
/approve

pkg/watcher/results/results.go
```diff
@@ -246,11 +247,44 @@ func copyKeys(in, out map[string]string) {
...
+	var endTime *timestamppb.Timestamp
+
+	switch obj := o.(type) {
+
+	case *pipelinev1.PipelineRun:
+		if obj.Status.CompletionTime != nil {
+			endTime = timestamppb.New(obj.Status.CompletionTime.Time)
+		}
+
+	case *pipelinev1.TaskRun:
```
khrm, line 291: ATM, missing type returns nil. We can have that explicit as the default. Like in the case of CustomRun.

pkg/watcher/results/results_test.go
```diff
@@ -630,3 +635,295 @@ func client(t *testing.T) *Client {
...
+				},
+				Status: pipelinev1.TaskRunStatus{
+					TaskRunStatusFields: pipelinev1.TaskRunStatusFields{
+						StartTime: &metav1.Time{Time: ts},
+					},
+				},
+			},
+			want: timestamppb.New(ts),
+		},
+		{
```
khrm, line 693: We can add a test case for CustomRun also. Probably a nil.

pkg/watcher/results/results_test.go
```diff
@@ -630,3 +635,295 @@ func client(t *testing.T) *Client {
...
+			name: "TaskRun without CompletionTime",
+			object: &pipelinev1.TaskRun{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: "tekton.dev/v1",
+					Kind:       "TaskRun",
+				},
+			},
+			want: nil,
+		},
+	}
```
khrm, line 847: Add a test case for CustomRun here also.

pkg/watcher/results/results.go
SergK: Done. Added explicit `default: return nil`. Also added a `CustomRun` case reading `status.StartTime/CompletionTime` - standalone CustomRuns are top-level records, so nil would have regressed their timestamps. `Matches getCompletionTime()` in the dynamic reconciler.

pkg/watcher/results/results_test.go
SergK: Done. Added CustomRun with/without StartTime — returns the actual timestamp, not nil, since CustomRun is now handled. Also added an unrecognized-type case for the default branch.

pkg/watcher/results/results_test.go
SergK: Done. Same coverage as above, plus CustomRun succeeded/failed cases in `TestEnsureResult_Timestamps.`

SergK:
@khrm  I've rebased the branch and resolved the comments in the second commit. Please let me know if you're expecting a squash on my side. thanks

PR #1306, opened 2026-04-20
fix(CLI): use ListRecords for old CLI log listing

divyansh42:
# Changes
The `logs list` command in the old CLI fails with `unknown service tekton.results.v1alpha2.Logs` because many
deployments do not register the v1alpha2 Logs gRPC service. When LOGS_API is disabled (default) or a log plugin is active (e.g. S3/GCS/Loki), only v1alpha3 Logs (GetLog) or no Logs service is exposed, causing the ListLogs RPC to fail.
Fix by switching to Results.ListRecords with a CEL filter for log record types (v1alpha2.Log and v1alpha3.Log). Record names are rewritten from records/ to logs/ format to preserve output compatibility. This avoids depending on the optional Logs gRPC service entirely. Also clean up cmd/tkn-results/main.go formatting and add
unit tests for the new helper functions.

Signed-off-by: divyansh42 <diagrawa@redhat.com>
Assisted-by: Claude Opus 4 (via Claude Code)

# Submitter Checklist

# Release Notes

```release-note
Fix `logs list` command failing with "unknown service tekton.results.v1alpha2.Logs" on deployments where the Logs gRPC service is not registered.
```

### DIFF
--- a/pkg/cli/dev/cmd/logs/list.go
+++ b/pkg/cli/dev/cmd/logs/list.go
@@ -18,6 +18,9 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/tektoncd/results/pkg/api/server/v1alpha2/log"
+	"github.com/tektoncd/results/pkg/api/server/v1alpha2/record"
+	"github.com/tektoncd/results/pkg/api/server/v1alpha2/result"
 	"github.com/tektoncd/results/pkg/cli/dev/flags"
 	"github.com/tektoncd/results/pkg/cli/dev/format"
 
@@ -34,16 +37,19 @@ func ListLogsCommand(params *flags.Params) *cobra.Command {
 		Short: "[To be deprecated] List Logs for a given Result",
 		Long:  "List Logs for a given Result. <result-name> is typically of format <namespace>/results/<parent-run-uuid>. '-' may be used in place of <parent-run-uuid> to query all Logs for a given parent.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			resp, err := params.LogsClient.ListLogs(cmd.Context(), &pb.ListRecordsRequest{
+			// Use Results.ListRecords (not Logs.ListLogs): many deployments omit tekton.results.v1alpha2.Logs.
+			req := &pb.ListRecordsRequest{
 				Parent:    args[0],
-				Filter:    opts.Filter,
+				Filter:    buildLogListFilter(opts.Filter),
 				PageSize:  opts.Limit,
 				PageToken: opts.PageToken,
-			})
+			}
+			resp, err := params.ResultsClient.ListRecords(cmd.Context(), req)
 			if err != nil {
 				fmt.Printf("List Logs: %v\n", err)
 				return err
 			}
+			rewriteLogRecordNames(resp.Records)
 			return format.PrintProto(os.Stdout, resp, opts.Format)
 		},
 		Args: cobra.ExactArgs(1),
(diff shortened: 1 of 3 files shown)


pkg/cli/dev/cmd/logs/list.go
```diff
@@ -62,3 +70,26 @@ func ListLogsCommand(params *flags.Params) *cobra.Command {
...
+}
+
+func rewriteLogRecordNames(records []*pb.Record) error {
+	for _, api := range records {
+		if api == nil {
+			continue
+		}
+		parent, resultName, recordName, err := record.ParseName(api.Name)
+		if err != nil {
+			return err
```
enarha: Returning an error here is sort of all or nothing approach. Wouldn't it be better if we log a warning, ignore the bad name and continue with the rest? Logging is trickier with CLI as we print that on the terminal for the user instead of writing it in a log, but personally I think it's still the more sensible approach. The warning/error can also be printed on the stderr.

pkg/cli/dev/cmd/logs/list_test.go
```diff
@@ -0,0 +1,50 @@
...
+	} {
+		got := buildLogListFilter(tc.user)
+		if diff := cmp.Diff(tc.want, got); diff != "" {
+			t.Errorf("buildLogListFilter(%q) diff (-want +got):\n%s", tc.user, diff)
+		}
+	}
+}
+
+func TestRewriteLogRecordNames(t *testing.T) {
+	in := "ns/results/res-1/records/rec-log"
```
enarha: Ideally we should always test with more than one record. I've seen code which only worked when exactly one entry was passed and had test to satisfy that "limitation". We should also mix a bad names with the good ones.

enarha (commented):
In addition to the code comments, did you try the functionality before and after the change on the same list of stored records? Are the records returned in the same order? Not sure it's critical, just trying to establish how backward compatible the change is.

pkg/cli/dev/cmd/logs/list.go
divyansh42: So do you mean to say output whatever list we got and for the rest log a error message?

pkg/cli/dev/cmd/logs/list_test.go
divyansh42: That's a good point. I will improve the tests

pkg/cli/dev/cmd/logs/list.go
enarha: No. What I meant is to filter out the bad names. As you loop on the list of records, if you get an error, print warning/error and then continue with the next record in the list.

khrm (commented):
/approve

pkg/cli/dev/cmd/logs/list.go
```diff
@@ -63,3 +70,32 @@ func ListLogsCommand(params *flags.Params) *cobra.Command {
...
+// buildLogListFilter restricts ListRecords to Log record types; optional user CEL is AND-combined.
+func buildLogListFilter(user string) string {
+	const logTypes = `(data_type == "results.tekton.dev/v1alpha3.Log" || data_type == "results.tekton.dev/v1alpha2.Log")`
+	if user == "" {
+		return logTypes
+	}
+	return fmt.Sprintf("(%s) && %s", user, logTypes)
+}
+
+func rewriteLogRecordNames(records []*pb.Record) error {
```
enarha: After the last change, we no longer return an error here, we always return nil, so better return nothing at all.

PR #1389, opened 2026-07-20
fix(cli): honor $KUBECONFIG and resolve exec credential auth

### LINKED ISSUE #1388: tkn-results CLI ignores $KUBECONFIG and does not honor kubeconfig exec credentials (OpenShift external OIDC / oc get-token)
# Expected Behavior

The CLI should authenticate to the Results API by **reusing the current kubeconfig context's credentials the way `oc`/`kubectl`/`tkn` do** — resolving the credential through client-go rather than reading a single static field (`rest.Config.BearerToken`) up front.

Specifically:

- **Honor `$KUBECONFIG`** (including multiple colon-separated, merged files) with the standard client-go precedence: `--kubeconfig` flag → `$KUBECONFIG` → `~/.kube/config`. This matches `oc`/`kubectl`/`tkn`.
- **Resolve the current context's bearer token via client-go**, honoring `exec` credential plugins — not only a static `token:`. A bearer token is not always a static string in the kubeconfig: it may be minted at request time by an `exec` plugin. On an OpenShift cluster configured for external OIDC, `oc login` writes a kubeconfig user with **no static `token:`**; `oc` itself is the credential plugin (`oc get-token`). `oc`/`kubectl`/`tkn` run this transparently and forward the resulting token, so `tkn-results` should too — automatically, with no manual `--token`.
- **Precedence:** explicit `--host`/`--token` (or `config set`) always override; otherwise fall back to the resolved kubeconfig token; only error if neither yields a usable credential.

This is consistent with how Results authenticates callers server-side: the API uses the Kubernetes **TokenReview + SubjectAccessReview** pattern (pulls `Authorization: Bearer <token>`, runs `TokenReview`, then `SubjectAccessReview` — [`pkg/api/server/v1alpha2/auth/rbac.go#L85-L144`](https://github.com/tektoncd/results/blob/main/pkg/api/server/v1alpha2/auth/rbac.go)). The token produced by `oc get-token` is exactly a TokenReview-able bearer token, so the server already accepts it — the only gap is the client not resolving it from the kubeconfig.

# Actual Behavior

When `tkn-results` builds its client from the kubeconfig:

1. **It ignores the `$KUBECONFIG` environment variable** — it hardcodes `~/.kube/config` and only respects an explicit `--kubeconfig` flag ([`pkg/cli/config/config.go#L62-L68`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/config/config.go#L62-L68), [`config.go#L404-L411`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/config/config.go#L404-L411)). `getRawKubeConfigLoader` builds `ClientConfigLoadingRules` with only `ExplicitPath`, bypassing the `clientcmd.NewDefaultClientConfigLoadingRules()` precedence chain that would consult `$KUBECONFIG`.

2. **It does not honor `exec` credential plugins** (`users[].user.exec`). It reads the *static* `BearerToken` off the resolved `rest.Config` ([`config.go#L126-L128`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/config/config.go#L126-L128), [`config.go#L388-L392`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/config/config.go#L388-L392)). For an `exec` credential, client-go populates `rest.Config.ExecProvider`, not `BearerToken`; the plugin only runs inside the transport during a live request. Because Results extracts and forwards the token itself instead of letting a client-go transport perform the request, the `exec` plugin is never executed and no token is obtained. As a result, `tkn-results` cannot authenticate to the Results API on external-OIDC OpenShift clusters without a manually supplied `--token`.

Both behaviors also diverge from another path in this same package, which already uses the env-aware loader ([`pkg/cli/common/params.go#L56`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/common/params.go#L56)) — suggesting the hardcoded `RecommendedHomeFile` in `NewConfig` was an oversight rather than intentional.

Verified with `tkn-results` **v0.19.0** (`go install github.com/tektoncd/results/cmd/tkn-results@latest`) and `oc` v4.22.4, against a real OpenShift cluster whose kubeconfig user is `oc get-token` (external OIDC, no static `token:`). Cluster name, tenant, client-id, user, and token are redacted below.

The kubeconfig user under test (external OIDC — no static `token:`):

```yaml
users:
- name: user@example.com/api-mycluster-example-com:6443
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1
      command: oc
      args:
      - get-token
      - --issuer-url=https://login.microsoftonline.com/<tenant>/v2.0
      - --client-id=<client-id>
      - --callback-address=127.0.0.1:0
      - --extra-scopes=email,profile
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

**1. `$KUBECONFIG` is ignored.** The working context lives *only* in `$KUBECONFIG`; the default `~/.kube/config` is empty. `oc` uses `$KUBECONFIG`; `tkn-results` reads the empty default file instead:

```console
$ export KUBECONFIG=/path/to/oidc-cluster.kubeconfig   # default ~/.kube/config is empty

$ oc config current-context
default/api-<cluster>:6443/<user>          # oc honors $KUBECONFIG

$ tkn-results config view
Error: context '' not found in kubeconfig   # tkn-results ignored $KUBECONFIG, read empty ~/.kube/config

$ tkn-results pipelinerun list -n default
Error: context '' not found in kubeconfig
```

(`tkn-results`' own `--namespace` help text even claims *"default: from $KUBECONFIG"*, but the value is never read.)

**2. `exec` credential (`oc get-token`) is not honored.** Bypassing (1) with an explicit `--kubeconfig`, using the **same** kubeconfig and the **same** exec-plugin token cache for both tools:

```console
# oc runs the exec plugin, obtains a token, authenticates:
$ oc --kubeconfig=/path/to/oidc-cluster.kubeconfig whoami
<user>@example.com

# tkn-results, same kubeconfig / same cache — extracts NO token from the exec credential:
$ tkn-results --kubeconfig=/path/to/oidc-cluster.kubeconfig config view
api-path: ""
apiVersion: results.tekton.dev/v1alpha2
host: ""
kind: Client
token: ""            # <-- empty: BearerToken never populated from the exec provider

$ tkn-results --kubeconfig=/path/to/oidc-cluster.kubeconfig pipelinerun list -n default
Error: invalid config. You may need to run the 'config set' command to configure the CLI: API server host not configured: host field is empty
```

The empty `token: ""` corresponds directly to `RESTConfig.BearerToken` being empty; because no token is resolved, host auto-detection (which must reach the cluster API) also fails with an empty `host`.

# Additional Info

**Root cause (permalinks pinned to `main` @ [`da051e6`](https://github.com/tektoncd/results/tree/da051e620e19bc0491963ef2a86ae996e82a92b9)):**

- `$KUBECONFIG` ignored: [`config.go#L62-L68`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/config/config.go#L62-L68) + [`config.go#L404-L411`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/config/config.go#L404-L411)
- `exec` credential not resolved: [`config.go#L126-L128`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/config/config.go#L126-L128) + [`config.go#L388-L392`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/config/config.go#L388-L392)
- Working env-aware loader already used elsewhere: [`common/params.go#L56`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/common/params.go#L56)
- Introduced in commit [`1eae7bb6`](https://github.com/tektoncd/results/commit/1eae7bb610077b6a97738226f2ebea88e9f90e62) ("[CLI enhancement] Add config command"), PR [#988](https://github.com/tektoncd/results/pull/988), first released in v0.16.0; still present on `main`.

**Suggested fix:**

1. In `NewConfig`, use `clientcmd.NewDefaultClientConfigLoadingRules()` (setting `ExplicitPath` only when `--kubeconfig` is given), matching `common/params.go`, so `$KUBECONFIG` is honored.
2. Obtain the effective bearer token via client-go's credential resolution instead of reading `rest.Config.BearerToken` directly, so the `exec` plugin (e.g. `oc get-token`) is actually run and its token forwarded. The transport plumbing is mostly there already (`LoadClientConfig` builds a `transport.Config` from `rc.TransportConfig()` and hands it to `transport.New()` at [`client/base.go#L63`](https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/pkg/cli/client/base.go#L63)). Static tokens keep working; `--token`/`--host` remain explicit overrides.

**Note on `--token`:** the sanctioned override is the current `--token` flag on the `pipelinerun`/`taskrun`/`config` commands (and `config set --token`) — not the legacy root flag `--authtoken` (`-t`), which was deprecated in [#1301](https://github.com/tektoncd/results/pull/1301) in favor of `config set --token`. But requiring users to hand-paste a token defeats the purpose of exec-based kubeconfig credentials; the plugin should be resolved automatically, with manual `--token` reserved for the disconnected / out-of-cluster case.

**Out of scope:** this issue is about **bearer tokens (static + `exec`/OIDC)**. A purely x509 (client-certificate) kubeconfig doesn't produce a bearer token, and the Results API authenticates via `TokenReview` (bearer-only) and never requests a client certificate — so cert-only kubeconfigs are a separate, server-side concern not addressed here.

**Also affects:** `opc results` (openshift-pipelines/opc), which vendors this CLI.

- Kubernetes distribution: OpenShift (external OIDC / "direct OIDC" issuer; kubeconfig user authenticates via `oc get-token` exec plugin)

  **Output of `oc version` (client):**

  ```
  Client Version: 4.22.4
  ```

- Tekton Results / CLI version:

  **`tkn-results` version:** v0.19.0 (`go install github.com/tektoncd/results/cmd/tkn-results@latest`); bug also present on `main` @ `da051e6`.

  Host OS/Go: macOS (darwin/arm64), Go 1.26.5.

dronenb:
# Changes

Fix CLI authentication that used its own kubeconfig handling instead of the standard client-go behavior used by `oc`/`kubectl`/`tkn`.

`tkn-results` built its Kubernetes client configuration from a `ClientConfigLoadingRules` with only an `ExplicitPath`, so `NewConfig` ignored the `$KUBECONFIG` environment variable and could not locate the current context when the credential lived in a kubeconfig referenced via `$KUBECONFIG`. On external-OIDC OpenShift clusters this meant `tkn-results` could not even find the context to authenticate against.

- **Honor `$KUBECONFIG`.** Switch to `clientcmd.NewDefaultClientConfigLoadingRules()` so the standard precedence is honored: an explicit `--kubeconfig` flag, then `$KUBECONFIG` (including multiple colon-separated, merged files), then `~/.kube/config`. An explicit `--kubeconfig` path still wins over the environment.
- **Don't pin short-lived credentials in `config set`.** The interactive Token prompt now pre-fills only for static (non-expiring) tokens. For dynamic credentials — exec credential plugins (e.g. `oc get-token` for OpenShift external OIDC), legacy auth-providers (oidc/gcp/azure), and token files — the prompt is left blank so a freshly-minted, short-lived token is never persisted into the kubeconfig extension, which would pin an expiring value and break later requests.
- Once the kubeconfig is located, exec credentials are resolved and forwarded on live requests by the existing client-go transport chain (`rest.Config.TransportConfig()` → `transport.New`), exactly the same way `oc`/`kubectl`/`tkn` work; no new token-resolution machinery is needed.
- Correct the `--kubeconfig` and `--namespace` flag help text to match the actual precedence/default behavior.
- Add unit coverage for kubeconfig precedence, `NewConfig`, static token handling, nil config handling, the dynamic-token prompt guard, and CLI flag help text.

Fixes #1388

# Submitter Checklist

# Release Notes

```release-note
tkn-results now honors $KUBECONFIG when loading kubeconfig, using the standard client-go precedence (--kubeconfig, then $KUBECONFIG, then ~/.kube/config). This fixes CLI authentication on clusters whose credentials live in a kubeconfig referenced by $KUBECONFIG, including OpenShift external-OIDC configs that rely on `oc get-token` exec credentials. Interactive `config set` no longer pre-fills/persists short-lived tokens from exec/auth-provider/token-file credentials, so the CLI keeps resolving them on each request.
```

### DIFF
--- a/pkg/cli/config/token.go
+++ b/pkg/cli/config/token.go
@@ -1,0 +1,97 @@
+package config
+
+import (
+	"errors"
+	"net/http"
+	"strings"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+)
+
+// resolveBearerToken returns the bearer token that client-go would attach to a
+// request built from the given rest.Config.
+//
+// A kubeconfig context does not always carry a static bearer token: the token
+// may instead be produced at request time by an exec credential plugin (e.g.
+// `oc get-token` for OpenShift external OIDC, `aws eks get-token`, etc.) or a
+// legacy auth-provider (oidc/gcp/azure). In those cases rest.Config.BearerToken
+// is empty and the credential is injected via the transport's WrapTransport
+// round-tripper chain instead.
+//
+// This helper drives that round-tripper chain once so the exec plugin /
+// auth-provider is actually invoked, and captures the resulting
+// "Authorization: Bearer <token>" header — the same token oc/kubectl/tkn would
+// send. If the config already has a static token, that is returned directly and
+// no plugin is run.
+//
+// It returns an empty string (and no error) when the context authenticates by
+// some means other than a bearer token (e.g. client certificate), since there
+// is no token to resolve in that case.
+func resolveBearerToken(rc *rest.Config) (string, error) {
+	if rc == nil {
+		return "", errors.New("nil REST config provided")
+	}
+
+	// Fast path: a static token is already present.
+	if rc.BearerToken != "" {
+		return rc.BearerToken, nil
+	}
+
+	tc, err := rc.TransportConfig()
+	if err != nil {
+		return "", err
+	}
+
+	// If the transport does not wrap requests with a credential provider
+	// (exec/auth-provider), there is no bearer token to resolve here.
+	if tc.WrapTransport == nil {
+		return "", nil
+	}
+
+	// Build the credential-aware round-tripper chain and run a single request
+	// through it. The terminal round-tripper never touches the network: it just
+	// captures whatever Authorization header the chain set.
+	capture := &authHeaderCapturingRoundTripper{}
+	rt, err := transport.HTTPWrappersForConfig(tc, capture)
+	if err != nil {
+		return "", err
+	}
+
+	// A minimal request; the host is irrelevant because capture short-circuits
+	// before any network I/O.
+	req, err := http.NewRequest(http.MethodGet, "https://tekton-results.local/", nil)
+	if err != nil {
+		return "", err
+	}
+	if _, err := rt.RoundTrip(req); err != nil {
+		return "", err
+	}
+
+	authz := capture.authorization
+	if authz == "" {
+		return "", nil
+	}
+	// Strip the "Bearer " scheme prefix if present.
+	if parts := strings.SplitN(authz, " ", 2); len(parts) == 2 && strings.EqualFold(parts[0], "Bearer") {
+		return parts[1], nil
+	}
+	return authz, nil
+}
+
+// authHeaderCapturingRoundTripper is a terminal http.RoundTripper that records
+// the Authorization header set by the wrapping credential round-trippers and
+// returns a stub response without performing any network I/O.
+type authHeaderCapturingRoundTripper struct {
+	authorization string
+}
+
+func (r *authHeaderCapturingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	r.authorization = req.Header.Get("Authorization")
+	return &http.Response{
+		StatusCode: http.StatusOK,
+		Header:     make(http.Header),
+		Body:       http.NoBody,
+		Request:    req,
+	}, nil
+}
--- a/pkg/cli/cmd/config/set.go
+++ b/pkg/cli/cmd/config/set.go
@@ -55,7 +55,9 @@ Usage Modes:
 
 Configuration Options:
 - Host: Tekton Results API server URL
-- Token: Bearer token (defaults to current kubeconfig token)
+- Token: Bearer token. Leave blank to resolve automatically from your kubeconfig
+  (recommended for exec/OIDC credentials, so it refreshes each request), or enter
+  a static token explicitly to persist it.
 - API Path: API endpoint path
 - TLS Settings: Certificate verification options
 
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -60,12 +60,9 @@ type config struct {
 //   - Config: A new Config instance if successful.
 //   - error: An error if any step in the configuration process fails.
 func NewConfig(p common.Params) (Config, error) {
-	kubeconfigPath := clientcmd.RecommendedHomeFile
-	if p.KubeConfigPath() != "" {
-		kubeconfigPath = p.KubeConfigPath()
-	}
-	// Load kubeConfig
-	cc := getRawKubeConfigLoader(kubeconfigPath)
+	// Load kubeConfig honoring the standard client-go precedence
+	// (--kubeconfig flag -> $KUBECONFIG -> ~/.kube/config).
+	cc := getRawKubeConfigLoader(p.KubeConfigPath())
 	ca := cc.ConfigAccess()
 	ac, err := cc.RawConfig()
 	if err != nil {
@@ -380,30 +377,51 @@ func (c *config) Host() string {
 	return url // Return the detected URL as default
 }
 
-// Token returns the bearer token from the REST configuration.
-// It returns an error if the REST configuration is not properly initialized.
+// Token returns the bearer token for the current kubeconfig context.
+//
+// If the context carries a static token it is returned directly. If the context
+// authenticates via an exec credential plugin (e.g. `oc get-token`) or a legacy
+// auth-provider, the plugin/provider is invoked to mint the token so it can be
+// forwarded to the Results API, matching oc/kubectl/tkn behavior. It returns an
+// error only if the REST configuration is not initialized or token resolution
+// fails.
 //
 // Returns:
-//   - any: The bearer token string if successful, or an error if the configuration is invalid.
+//   - any: The bearer token string if successful, an empty string if the
+//     context authenticates by non-token means, or an error on failure.
 func (c *config) Token() any {
 	if c.RESTConfig == nil {
 		return fmt.Errorf("REST configuration is not initialized")
 	}
-	return c.RESTConfig.BearerToken
+	token, err := resolveBearerToken(c.RESTConfig)
+	if err != nil {
+		return err
+	}
+	return token
 }
 
-// getRawKubeConfigLoader creates and returns a clientcmd.ClientConfig based on the provided kubeconfig path.
-// This function is equivalent to ToRawKubeConfigLoader() and is used to load the kubeconfig file.
+// getRawKubeConfigLoader creates and returns a clientcmd.ClientConfig using the
+// standard client-go loading rules. This function is equivalent to
+// ToRawKubeConfigLoader() and honors the usual kubeconfig precedence:
+// the --kubeconfig flag (explicitPath), then the $KUBECONFIG environment
+// variable (which may list multiple, colon-separated files that are merged),
+// and finally ~/.kube/config.
 //
 // Parameters:
-//   - kubeconfigPath: A string representing the path to the kubeconfig file.
+//   - explicitPath: An optional explicit kubeconfig path (e.g. from --kubeconfig).
+//     When empty, $KUBECONFIG / the default home file are used.
 //
 // Returns:
-//   - clientcmd.ClientConfig: A non-interactive deferred loading client configuration
-//     that uses the specified kubeconfig path and default overrides.
-func getRawKubeConfigLoader(kubeconfigPath string) clientcmd.ClientConfig {
-	// Set explicit path for kubeconfig
-	loadingRules := &clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeconfigPath}
+//   - clientcmd.ClientConfig: A non-interactive deferred loading client configuration.
+func getRawKubeConfigLoader(explicitPath string) clientcmd.ClientConfig {
+	// Use the default loading rules so that $KUBECONFIG is honored, matching
+	// the behavior of oc/kubectl/tkn and the rest of this package (see
+	// common.Params). An explicit --kubeconfig path, when provided, takes
+	// precedence over $KUBECONFIG.
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	if explicitPath != "" {
+		loadingRules.ExplicitPath = explicitPath
+	}
 	configOverrides := &clientcmd.ConfigOverrides{}
 
 	// Return the clientcmd.ClientConfig (equivalent to ToRawKubeConfigLoader)
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -60,12 +60,9 @@ type config struct {
 //   - Config: A new Config instance if successful.
 //   - error: An error if any step in the configuration process fails.
 func NewConfig(p common.Params) (Config, error) {
-	kubeconfigPath := clientcmd.RecommendedHomeFile
-	if p.KubeConfigPath() != "" {
-		kubeconfigPath = p.KubeConfigPath()
-	}
-	// Load kubeConfig
-	cc := getRawKubeConfigLoader(kubeconfigPath)
+	// Load kubeConfig honoring the standard client-go precedence
+	// (--kubeconfig flag -> $KUBECONFIG -> ~/.kube/config).
+	cc := getRawKubeConfigLoader(p.KubeConfigPath())
 	ca := cc.ConfigAccess()
 	ac, err := cc.RawConfig()
 	if err != nil {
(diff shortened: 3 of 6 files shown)


dronenb:
Addressed Copilot review comments in `0a155407`:

- `resolveBearerToken` now always runs client-go HTTP auth wrappers after the static-token fast path, so `BearerTokenFile` / kubeconfig `tokenFile` credentials resolve even when `WrapTransport` is nil. Added `TestResolveBearerTokenFile` coverage.
- Non-Bearer `Authorization` schemes now resolve to an empty token instead of being returned verbatim, and the test expectation was updated accordingly.

pkg/cli/config/token.go
```diff
@@ -0,0 +1,92 @@
...
+	// before any network I/O.
+	req, err := http.NewRequest(http.MethodGet, "https://tekton-results.local/", nil)
+	if err != nil {
+		return "", err
+	}
+	if _, err := rt.RoundTrip(req); err != nil {
+		return "", err
+	}
+
+	authz := capture.authorization
```
aThorp96: I think this can be simplified a bit by accessing the attributes directly, something like the below instead of the `authHeaderCapturingRoundTripper` 

```suggestion
	type noopRoundTripper struct{}
	func (_ *noopRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
		return &http.Response{Request: req}
	}
	rt, err := transport.HTTPWrappersForConfig(tc, &noopRoundTripper{})
	if err != nil {
		return "", err
	}
	
	req := &http.Req{}
	 if _, err := rt.RoundTrip(req); err != nil {
		return "", err
	}

	authz := req.Header.Get("Authorization")
```

aThorp96 (commented):
This is great, thanks! Clever idea to extract token injected by middleware as well. I have one suggestion for simplifying the token resolution logic. It should also reduce the required tests a bit

dronenb:
Addressed @aThorp96 review feedback in `bf61273c` by removing the dedicated `authHeaderCapturingRoundTripper` and using a small terminal `roundTripperFunc` instead. One implementation note: client-go bearer auth can clone the request, so the resolver reads `Authorization` from `resp.Request` after the wrapper chain runs rather than from the original request. Validation: `go test ./pkg/cli/...`, `./test/presubmit-tests.sh --unit-tests`, and `make all` all pass.

dronenb:
I will do some testing with this against our Tekton clusters and mark as ready for review when I'm comfortable things are good to go

aThorp96:
> One implementation note: client-go bearer auth can clone the request, so the resolver reads Authorization from resp.Request after the wrapper chain runs rather than from the original request.

Ah, good catch. I was wondering about that possibility as well but wasn't sure.

enarha:
> I will do some testing with this against our Tekton clusters and mark as ready for review when I'm comfortable things are good to go

Thank you for this PR. Can you please remove the draft/not ready indicator.

enarha:
@divyansh42 PTAL when you have a chance.

dronenb:
I think I identified an issue with how I was testing this, so I marked it as draft. Will validate today or tomorrow...

divyansh42 (commented):
@dronenb Quick question on the config set interactive prompt: for exec/OIDC credentials (e.g. oc get-token), it now pre-fills the "Token" field with a live, short-lived token instead of leaving it blank. If a user just hits Enter, that token gets saved permanently in Extension.Token, and future commands will keep using that saved value instead of asking oc for a fresh one — so things would start failing again once it expires, until the user reruns config set.

Is that the intended behavior, or should the prompt stay blank by default for non-static credentials so the CLI keeps auto-refreshing the token each time?

divyansh42:
> I think I identified an issue with how I was testing this, so I marked it as draft. Will validate today or tomorrow...

@dronenb just wanted to check, if you were able to validate it?

dronenb:
> @dronenb Quick question on the config set interactive prompt: for exec/OIDC credentials (e.g. oc get-token), it now pre-fills the "Token" field with a live, short-lived token instead of leaving it blank. If a user just hits Enter, that token gets saved permanently in Extension.Token, and future commands will keep using that saved value instead of asking oc for a fresh one — so things would start failing again once it expires, until the user reruns config set.
> 
> Is that the intended behavior, or should the prompt stay blank by default for non-static credentials so the CLI keeps auto-refreshing the token each time?

I have addressed this, PTAL


> @dronenb just wanted to check, if you were able to validate it?

Just validated, should be g2g. I have rebased as well.

divyansh42 (approved):
Thanks @dronenb!
Could you please squash the commits into one?

dronenb:
> Thanks @dronenb! Could you please squash the commits into one?

Thanks, @divyansh42! I have squashed the commits and rebased, should be g2g!

divyansh42:
> > Thanks @dronenb! Could you please squash the commits into one?
> 
> Thanks, @divyansh42! I have squashed the commits and rebased, should be g2g!

/cc @enarha @khrm

dronenb:
@divyansh42 @enarha @khrm any update on this? Anything needed from my end? Let me know! 🙂

enarha:
> @divyansh42 @enarha @khrm any update on this? Anything needed from my end? Let me know! 🙂

Not for now. I want to perform a test on OpenShift cluster to verify if a smaller fix isn't sufficient. Please give me a day or max two to check that.

enarha:
Thanks for this — the fix works, and I verified it end-to-end against a live OpenShift (ROSA) cluster. I ran a test to isolate *which* part of the change actually carries the exec-credential flow, and I think it points to a simplification opportunity.

**What I tested**

I built a variant of this branch with **only the `$KUBECONFIG` precedence fix kept** and the token-resolution part reverted:
- deleted `pkg/cli/config/token.go` (`resolveBearerToken` / `usesDynamicToken`)
- reverted `Token()` to `return c.RESTConfig.BearerToken`
- reverted `Set()` to `token := c.Token()` and removed `tokenPromptDefault()`
- kept `getRawKubeConfigLoader` using `NewDefaultClientConfigLoadingRules()` and the `NewConfig` change

I pointed it at the cluster with a synthetic **exec-based kubeconfig** (an exec plugin standing in for `oc get-token`, with **no static `token:`**), and confirmed the CLI's stored config token was empty — so any successful auth could only come from the exec plugin being invoked at request time.

| Test | Exec plugin returns | Outcome |
|------|--------------------|---------|
| Main | valid token | :white_check_mark: `pipelinerun list` succeeded |
| Control A | bogus token | :x: `401 permission denied` (auth is enforced) |
| Control B | valid token restored | :white_check_mark: succeeded again |

So with `token.go` removed entirely, the exec-based OpenShift flow still works. The exec credential is resolved and forwarded by the **pre-existing transport chain**: `LoadClientConfig` -> `rc.TransportConfig()` already invokes `ExecProvider` the same way `oc`/`kubectl` do, and that code is untouched by this PR. The change that actually unblocks the flow is the `$KUBECONFIG` fix (it lets `NewConfig` locate the context at all).

**Suggested simplification**

Tracing the call graph, the exec-resolution machinery in `token.go` doesn't appear to be reachable on any real path:
- `resolveBearerToken` / `Token()` are only called from `tokenPromptDefault`.
- `tokenPromptDefault` returns `""` for any dynamic credential (`usesDynamicToken` -> exec / auth-provider / token-file) **before** calling `Token()`.
- So `resolveBearerToken` is only reached for non-dynamic contexts, where either `BearerToken` is already set (fast path returns it directly) or there's no credential (returns `""`).

In other words, the round-tripper logic that drives the exec plugin to capture the `Authorization` header is only invoked for exactly the cases that never need it. The genuinely useful behavior here — not pre-filling/persisting a short-lived token during `config set` — is provided by `usesDynamicToken` alone.

Consider dropping `resolveBearerToken` and its round-tripper, keeping `usesDynamicToken` for the `config set` prompt guard, and reverting `Token()` to returning `RESTConfig.BearerToken`. That shrinks the PR and keeps the observed behavior identical.

Either way, I'd also tweak the description/release note: the exec-token forwarding on live requests already worked via client-go's transport chain, so it's really the `$KUBECONFIG` fix doing that work — worth stating so the intent is clear for future maintainers.

Not blocking — behavior is correct as-is; this is about reducing surface area.

dronenb:
@enarha thanks for the thorough review! I have made changes to reduce the surface while preserving the functionality per your analysis.

PR #1411, opened 2026-07-31
fix(api,watcher): cascade delete Results on namespace deletion 

pratap0007:
**fix(api,watcher): cascade delete Results on namespace deletion and map Loki/Splunk HTTP errors to proper gRPC codes**

When a namespace is deleted, Results API records persisted in Postgres but log retrieval failed with a generic 500 Internal Error. The root cause was twofold:

1. The watcher had no Namespace informer, so Results/Records were never cleaned up when a namespace was deleted, leaving orphaned data.

2. The log plugin mapped ALL non-200 HTTP responses from Loki/Splunk to codes.Internal with a generic error message. When Loki returned 403 Forbidden for a deleted namespace tenant, this surfaced as an unhelpful "Error fetching log data (code: 500)" to the user.

This commit addresses both issues:

**Namespace cascade deletion (Fix 1):**
- Add a new Namespace reconciler (pkg/watcher/reconciler/namespace/) that watches for namespace deletion events via the Kubernetes informer
- On namespace deletion or DeletionTimestamp being set, the reconciler paginates through ListResults and calls DeleteResult for each result in that namespace. DB-level cascade (OnDelete:CASCADE) handles Records
- Register the new controller in the watcher's main.go

**HTTP-to-gRPC error mapping (Fix 2):**
- Add httpStatusToGRPCCode() helper mapping HTTP 400->InvalidArgument, 401->Unauthenticated, 403->PermissionDenied, 404->NotFound, 429->ResourceExhausted, and default->Internal
- Apply mapping to Loki and both Splunk error paths, including the HTTP status code in the error message for debuggability
- Fix response body leak in getLokiLogs (missing defer resp.Body.Close)

Resolves: SRVKP-12883

/kind bug

# Changes

# Submitter Checklist

# Release Notes

```release-note
Added a namespace reconciler to the watcher to handle namespace deletions. When a namespace is deleted, all associated Results are removed through the API, and Records are automatically cleaned up by database foreign key constraints, preventing orphaned data.

```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -89,6 +89,8 @@ Handles all data mutations and queries.
 **Watcher** (`cmd/watcher`, `pkg/watcher/`): Kubernetes controller that watches
 TaskRun, PipelineRun, and CustomRun resources. Creates or updates corresponding
 Records via the Results API. Annotates original CRDs with result identifiers.
+Also watches Namespace deletions and cascade-deletes associated Results/Records
+via a dedicated namespace reconciler.
 
 **Retention Policy Agent** (`cmd/retention-policy-agent`, `pkg/retention/`):
 Deletes old data from the database based on configured retention policies.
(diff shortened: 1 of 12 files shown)


khrm (commented):
Tbh, it makes sense to persist data for deleted namespace. It's an issue from Loki side that it's not persisting data.

khrm (commented):
Are we sure that this issue exist with Splunk. Also, does this issue exist with only OpenShift Loki or normal Loki also.

enarha:
> Tbh, it makes sense to persist data for deleted namespace. It's an issue from Loki side that it's not persisting data.

I agree. We should not simply purge the data for deleted namespace. The namespace could be deleted by mistake and easily re-created, but by that time the archived data would be gone. Furthermore, I think this should not be the watcher responsibility and I don't like this extra delete permission added to it. Instead that could be made responsibility of the retention-policy-agent (it already has this delete permission and that's its primary job) which can be enabled by the user with some flag e.g. `purge-data-for-deleted-namespace` and probably make that opt-in initially . As currently implemented, this is also a breaking change.

Providing proper error message is important though.

pratap0007:
> Are we sure that this issue exist with Splunk. Also, does this issue exist with only OpenShift Loki or normal Loki also.

It has an issue with Loki, and I haven't checked it with Splunk

PR #1433, opened 2026-08-09
chore(1432): fix golangci-lint/go version mismatch

### LINKED ISSUE #1432: CI is failing on golangci-lint version mismatch
# Expected Behavior
golangci-lint is passing without errors
# Actual Behavior
golangci-lint is failing with `Error: can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.5)`

1. Submit the PR
2. Wait for CI linter pipeline to get triggered
3. golangci-lint is failing with `Error: can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.5)`

olegbet:
Assisted-by: Claude

# Changes
**:warning:This is a temporary workaround to unblock the CI. Should be removed after workflow is fixed**
Adds `install-mode: "goinstall"` to golangci-lint workflow to compile golangci-lint from source for Go 1.26+ compatibility
Closes: #1432

/kind bug
# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -24,6 +24,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
-          install-mode: "goinstall"
+          install-mode: "goinstall" #This is a temporary workaround to unblock the CI. Should be removed after workflow is fixed
           version: v2.12.2
           args: --new-from-merge-base=origin/${{ github.base_ref }} --timeout=10m
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -24,5 +24,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
+          install-mode: "goinstall"
           version: v2.12.2
           args: --new-from-merge-base=origin/${{ github.base_ref }} --timeout=10m


.github/workflows/golangci-lint.yaml
```diff
@@ -24,5 +24,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
+          install-mode: "goinstall"
```
enarha: This is an easy workaround to unblock the CI, but I see it as a temporary solution as the docs recommends against it. Please add a comment above it saying this is a temporary solution and should be removed once we can upgrade to version newer than v9.3.0. This way we won't miss it when the new version is available and we update to it.

olegbet: Added a comment to the PR description

enarha: I actually meant comment in the code, above that line so we know what to clean up after we bump the github action version. Comment on the PR won't be very helpful in that regard.

olegbet: Addressed in commit 00117bf84b9bbf93ca2d1eed147a005973c673f0

enarha (commented):
/lgtm
/approve

divyansh42:
@olegbet @enarha can't we configure CI jobs in other branches to use the workflow file from the main so that we have a single place to maintain those workflows.
Something like:

```
jobs:
  ci:
    uses: org/repo/.github/workflows/ci.yml@main
```

enarha:
> @olegbet @enarha can't we configure CI jobs in other branches to use the workflow file from the main so that we have a single place to maintain those workflows. Something like:
> 
> ```
> jobs:
>   ci:
>     uses: org/repo/.github/workflows/ci.yml@main
> ```

Sounds like a good idea, for this workflow at least. For bigger and more complex workflows we might need to rewrite and parametrize the job.

PR #1434, opened 2026-08-11
watcher and api-server horizontal scaling

olegbet:
# Changes
- added Watcher pods bucket-sharding scaling swithched on in Watcher's gRPC config ;
- deprecated blocking grpc.DialContext client connection function has been replaced with newer grpc.NewClient function. grpc.WithBlock has been removed  and replaced with grpc.WithConnectParams(connectParams) to use new scaling parameters;
- added StatefulSet overlay `tekton-results-watcher` with 3 replicas to be used to scale Watcher pods instead of Deployment ;
- added Deployment overlay `tekton-results-api`  scaled to 3 replicas;
- Headless service `tekton-results-api-service-headless` has been added with `clusterIP: None` for load balancing to be able to use DNS resolution;
- TLS certificate generation script added for the new headless service;
- Deployment scripts modified to deploy a new horizontally scaled configuration;
- Integration and functional tests added;
- Documentation updates about horizontal scaling;

# Submitter Checklist

# Release Notes

```release-note
Added a horizontal scaling functionality for Watcher / API Server high availability.
Added the high availability kustomize overlay to deploy the new horizontaly scaled configuration.
The Watcher and API server can now be deployed scaled to multiple pods and they will use DNS-based round-robin gRPC load balancing to manage workload properly.
By default the service will be deployed the single replica way but the process how to deploy the scaled configuration can be found in documentation.
```

Assisted-by: Claude

### DIFF
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -37,6 +37,7 @@ import (
 	v1alpha2pb "github.com/tektoncd/results/proto/v1alpha2/results_go_proto"
 	"golang.org/x/oauth2"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/credentials/oauth"
@@ -59,6 +60,7 @@ const (
 	// This is a fixed path which does not contain a hard-coded secret or credential
 	podTokenPath             = "/var/run/secrets/kubernetes.io/serviceaccount/token" //nolint:gosec
 	finalizerRequeueInterval = 10 * time.Second
+	defaultServiceConfig     = `{"loadBalancingConfig":[{"round_robin":{}}]}`
 )
 
 var (
@@ -101,7 +103,7 @@ func main() {
 
 	ctx = filteredinformerfactory.WithSelectors(ctx, "app.kubernetes.io/name")
 
-	conn, err := connectToAPIServer(ctx, *apiAddr, *authMode)
+	conn, err := connectToAPIServer(*apiAddr, *authMode)
 	if err != nil {
 		log.Fatalf("did not connect: %v", err)
 	}
@@ -178,7 +180,7 @@ func main() {
 	)
 }
 
-func connectToAPIServer(ctx context.Context, apiAddr string, authMode string) (*grpc.ClientConn, error) {
+func connectToAPIServer(apiAddr string, authMode string) (*grpc.ClientConn, error) {
 	// Load TLS certs
 	certs, err := loadCerts()
 	if err != nil {
@@ -186,9 +188,17 @@ func connectToAPIServer(ctx context.Context, apiAddr string, authMode string) (*
 	}
 	cred := credentials.NewClientTLSFromCert(certs, "")
 
-	opts := []grpc.DialOption{
-		grpc.WithBlock(),
+	connectParams := grpc.ConnectParams{
+		Backoff:           backoff.DefaultConfig,
+		MinConnectTimeout: 1 * time.Minute,
 	}
+
+	opts := []grpc.DialOption{
+		grpc.WithDefaultServiceConfig(defaultServiceConfig),
+		grpc.WithConnectParams(connectParams),
+		grpc.WithDefaultCallOptions(grpc.WaitForReady(true)),
+	}
+
 	// Add in additional credentials to requests if desired.
 	switch authMode {
 	case "google":
@@ -215,9 +225,7 @@ func connectToAPIServer(ctx context.Context, apiAddr string, authMode string) (*
 	}
 
 	log.Printf("dialing %s...\n", apiAddr)
-	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
-	defer cancel()
-	return grpc.DialContext(ctx, apiAddr, opts...)
+	return grpc.NewClient(apiAddr, opts...)
 }
 
 func loadCerts() (*x509.CertPool, error) {
--- a/cmd/watcher/main_test.go
+++ b/cmd/watcher/main_test.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2026 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"encoding/json"
+	"testing"
+
+	"google.golang.org/grpc/balancer/roundrobin"
+)
+
+func TestDefaultServiceConfig(t *testing.T) {
+	var config struct {
+		LoadBalancingConfig []map[string]interface{} `json:"loadBalancingConfig"`
+	}
+
+	if err := json.Unmarshal([]byte(defaultServiceConfig), &config); err != nil {
+		t.Fatalf("defaultServiceConfig is not valid JSON: %v", err)
+	}
+
+	if len(config.LoadBalancingConfig) == 0 {
+		t.Fatal("defaultServiceConfig must contain at least one load balancing policy")
+	}
+
+	foundRoundRobin := false
+	for _, policy := range config.LoadBalancingConfig {
+		if _, ok := policy[roundrobin.Name]; ok {
+			foundRoundRobin = true
+			break
+		}
+	}
+
+	if !foundRoundRobin {
+		t.Errorf("defaultServiceConfig must include the round_robin load balancing policy, got: %v", config.LoadBalancingConfig)
+	}
+}
--- a/config/components/horizontal-scaling/watcher-statefulset.yaml
+++ b/config/components/horizontal-scaling/watcher-statefulset.yaml
@@ -1,0 +1,118 @@
+# Copyright 2026 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: watcher
+  labels:
+    app.kubernetes.io/name: tekton-results-watcher
+spec:
+  serviceName: watcher-headless
+  replicas: 3
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: tekton-results-watcher
+  template:
+    metadata:
+      annotations:
+        cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
+      labels:
+        app.kubernetes.io/name: tekton-results-watcher
+        app: tekton-results-watcher
+    spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                - key: kubernetes.io/os
+                  operator: NotIn
+                  values:
+                  - windows
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
+              labelSelector:
+                matchLabels:
+                  app.kubernetes.io/name: tekton-results-watcher
+              topologyKey: kubernetes.io/hostname
+            weight: 100
+      serviceAccountName: watcher
+      containers:
+        - name: watcher
+          image: ko://github.com/tektoncd/results/cmd/watcher
+          args:
+            - -api_addr
+            - $(TEKTON_RESULTS_API_SERVICE)
+            - -auth_mode
+            - $(AUTH_MODE)
+          env:
+            - name: SYSTEM_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: CONFIG_LOGGING_NAME
+              value: tekton-results-config-logging
+            - name: CONFIG_LEADERELECTION_NAME
+              value: tekton-results-config-leader-election
+            - name: CONFIG_OBSERVABILITY_NAME
+              value: tekton-results-config-observability
+            - name: METRICS_DOMAIN
+              value: tekton.dev/results
+            - name: TEKTON_RESULTS_API_SERVICE
+              value: dns:///tekton-results-api-service-headless.tekton-pipelines.svc.cluster.local:8080
+            - name: AUTH_MODE
+              value: token
+            - name: KUBERNETES_MIN_VERSION
+              value: "v1.28.0"
+            - name: STATEFUL_CONTROLLER_ORDINAL
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: STATEFUL_SERVICE_NAME
+              value: tekton-results-watcher-headless
+          ports:
+            - name: metrics
+              containerPort: 9090
+            - name: profiling
+              containerPort: 8008
+          volumeMounts:
+            - name: tls
+              mountPath: "/etc/tls"
+              readOnly: true
+          securityContext:
+            seccompProfile:
+              type: RuntimeDefault
+            runAsNonRoot: true
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
+      volumes:
+        - name: tls
+          secret:
+            secretName: tekton-results-tls
+
+---
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: watcher-pdb
+spec:
+  maxUnavailable: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: tekton-results-watcher
--- a/docs/horizontal-scaling.md
+++ b/docs/horizontal-scaling.md
@@ -1,0 +1,298 @@
+# Horizontal Scaling
+
+Tekton Results supports horizontal scaling for both the API server and the watcher to provide high availability and increased throughput.
+
+## Overview
+
+Horizontal scaling enables Tekton Results to:
+
+- **Increase availability** - Multiple replicas eliminate single points of failure
+- **Distribute load** - Workload is partitioned across watcher replicas and API requests are load-balanced across API replicas
+- **Scale throughput** - More replicas handle more concurrent reconciliations and API requests
+
+The horizontal scaling configuration deploys:
+
+- **3 API server replicas** behind a headless Service for DNS-based gRPC load balancing
+- **3 watcher replicas** as a StatefulSet with bucket-based workload partitioning
+- **3 buckets** for consistent hashing of Tekton resources across watchers
+
+## Architecture
+
+### API Server Scaling
+
+The API server scales horizontally using a standard Kubernetes Deployment with multiple replicas. Load balancing is achieved through DNS-based gRPC client-side load balancing.
+
+**Components:**
+
+- **Deployment** - `tekton-results-api` with 3 replicas
+- **Headless Service** - `tekton-results-api-service-headless` with `clusterIP: None`
+- **ClusterIP Service** - `tekton-results-api-service` for non-HA clients (preserved for backwards compatibility)
+
+**How it works:**
+
+1. The headless Service has `clusterIP: None`, causing DNS queries to return all pod IP addresses instead of a single virtual IP
+2. gRPC clients use the `dns:///` scheme to trigger DNS-based endpoint resolution
+3. The `round_robin` load balancing policy distributes requests evenly across all resolved endpoints
+4. Each watcher maintains a single gRPC connection that multiplexes requests to all API replicas via HTTP/2 streams
+
+**Why headless Service:**
+
+Standard Kubernetes ClusterIP Services load-balance at the TCP connection level. Since gRPC uses long-lived HTTP/2 connections, all requests from a single watcher would go to the same API pod. The headless Service enables DNS to return all pod IPs, allowing gRPC's client-side load balancer to distribute individual requests across all API replicas.
+
+### Watcher Scaling
+
+The watcher scales horizontally using a StatefulSet with bucket-based workload partitioning provided by Knative's leader-aware controllers.
+
+**Components:**
+
+- **StatefulSet** - `tekton-results-watcher` with 3 replicas (replaces the base Deployment)
+- **Headless Service** - `tekton-results-watcher-headless` for StatefulSet network identity
+- **ConfigMap** - `config-leader-election` with `buckets: "3"`
+
+**How it works:**
+
+1. Knative's controller framework partitions the reconciliation key space into buckets using consistent hashing
+2. Each StatefulSet replica is assigned one or more buckets based on its ordinal (derived from pod name)
+3. When a watcher observes a TaskRun, PipelineRun, or CustomRun, it computes the bucket hash from the resource's namespace/name
+4. The watcher only reconciles the resource if the computed bucket matches one of its assigned buckets
+5. Resources are deterministically assigned to exactly one watcher, eliminating duplicate reconciliation
+
+Note: Deprecated blocking grpc.DialContext client connection function has been replaced with newer grpc.NewClient function. 
+grpc.WithBlock is incompatible with the new scaling functionality and has been removed and replaced with 
+grpc.WithConnectParams(connectParams) to use new scaling parameters;
+
+
+**Bucket assignment example:**
+
+With 3 replicas and 3 buckets:
+- `tekton-results-watcher-0` reconciles bucket 0
+- `tekton-results-watcher-1` reconciles bucket 1
+- `tekton-results-watcher-2` reconciles bucket 2
+
+A TaskRun named `default/my-task` hashes to bucket 1, so only `watcher-1` reconciles it.
+
+**StatefulSet vs Deployment:**
+
+StatefulSets provide stable network identities and ordinals that Knative uses for bucket assignment. The `STATEFUL_CONTROLLER_ORDINAL` environment variable (set by Knative injection) identifies which replica this is, enabling the bucket-to-replica mapping.
+
+## Configuration
+
+### Deploying with Kustomize
+
+The horizontal scaling feature is packaged as a Kustomize component at `config/components/horizontal-scaling`.
+
+**Example overlay:**
+
+```yaml
+# config/overlays/ha-local-db/kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: tekton-pipelines
+resources:
+  - ../../base
+components:
+  - ../../components/local-db
+  - ../../components/horizontal-scaling
+  - ../../components/metadata
+```
+
+**Deploy:**
+
+```bash
+# Generate TLS certificate with both service names
+./config/components/horizontal-scaling/generate-tls-cert.sh
+
+# Deploy with kustomize and ko
+kubectl kustomize config/overlays/ha-local-db | ko apply -f -
+```
+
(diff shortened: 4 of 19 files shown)


khrm (commented):
Can you please remove reference to internal jira in the title?

config/components/horizontal-scaling/README.md
```diff
@@ -0,0 +1,241 @@
+# Horizontal Scaling Component
+
```
khrm: Nit: Can we move this to doc folder.

enarha:
Just to confirm, this is still WIP, right? If so, please indicate it on the PR e.g. prefix the PR subject with `WIP: `. This is also a significant change adding new functionality, so the `Release Notes` section of the PR description should include a short release note. The release notes are used when we create a new release to inform the users which changes are incorporated in that release.

cmd/watcher/main_test.go
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright 2020 The Tekton Authors
```
enarha:
```suggestion
Copyright 2026 The Tekton Authors
```

olegbet:
> Can you please remove reference to internal jira in the title?

Should I create an upstream github issue for this feature?

enarha:
> > Can you please remove reference to internal jira in the title?
> 
> Should I create an upstream github issue for this feature?

IMO that's not needed, just for the sake of having one. The GitHub issue is a way for the users and contributors to communicate issues and features requests. This one is almost done, so no really need for it.

enarha:
Please also add ha-base-only overlay (the same as the one you added without the local-db component)

config/components/horizontal-scaling/watcher-statefulset.yaml
```diff
@@ -0,0 +1,107 @@
...
+            - name: CONFIG_LEADERELECTION_NAME
+              value: tekton-results-config-leader-election
+            - name: CONFIG_OBSERVABILITY_NAME
+              value: tekton-results-config-observability
+            - name: METRICS_DOMAIN
+              value: tekton.dev/results
+            - name: TEKTON_RESULTS_API_SERVICE
+              value: dns:///tekton-results-api-service-headless.tekton-pipelines.svc.cluster.local:8080
+            - name: AUTH_MODE
+              value: insecure
```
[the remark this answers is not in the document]
olegbet: Fixed in commit 96622609d7010dde0f621c639a8069f44b1012fc

config/components/horizontal-scaling/README.md
olegbet: Fixed in commit 7950215e6ec83490ca9770f2e216500ddc71e537

config/components/horizontal-scaling/README.md
```diff
@@ -0,0 +1,241 @@
...
+# 2. Deploy with kustomize and ko
+kubectl kustomize config/overlays/ha-local-db | ko apply -f -
+```
+
+## Operational Notes
+
+### Scaling Behavior
+
+- **Scale down safety**: Headless Service DNS only returns Ready pod IPs. gRPC handles GOAWAY frames for graceful connection draining.
+- **Rollout**: Knative's leader-aware controllers promote/demote bucket ownership via informer resync during pod updates.
```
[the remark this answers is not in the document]
olegbet: Fixed in commit 7950215e6ec83490ca9770f2e216500ddc71e537

config/components/horizontal-scaling/watcher-statefulset.yaml
```diff
@@ -0,0 +1,107 @@
...
+            - name: CONFIG_LOGGING_NAME
+              value: tekton-results-config-logging
+            - name: CONFIG_LEADERELECTION_NAME
+              value: tekton-results-config-leader-election
+            - name: CONFIG_OBSERVABILITY_NAME
+              value: tekton-results-config-observability
+            - name: METRICS_DOMAIN
+              value: tekton.dev/results
+            - name: TEKTON_RESULTS_API_SERVICE
+              value: dns:///tekton-results-api-service-headless.tekton-pipelines.svc.cluster.local:8080
```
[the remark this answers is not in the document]
olegbet: **false-positive** as per userstory: no resolver code changes needed — this is why the standard "headless Service + round_robin" Kubernetes gRPC pattern only needs this one service-config change

olegbet:
> Please also add ha-base-only overlay (the same as the one you added without the local-db component)

Done

docs/horizontal-scaling.md
```diff
@@ -0,0 +1,291 @@
...
+**Safe scale-down:**
+
+Kubernetes deletes StatefulSet pods in reverse ordinal order (2, 1, 0). Knative promotes remaining replicas to take ownership of orphaned buckets via informer resync. No reconciliations are lost during scale-down.
+
+### Rolling Updates
+
+When updating the watcher StatefulSet (new image, config change):
+
+1. Kubernetes performs a rolling update, replacing one pod at a time
+2. Knative demotes the terminating pod's buckets and promotes a different replica
```
enarha, line 267: I believe this is wrong. The bucket is left without an owner until the statefulset creates the replica with the respective ordinal.

docs/horizontal-scaling.md
```diff
@@ -0,0 +1,291 @@
...
+3. The new pod starts, cache syncs, and takes ownership of assigned buckets
```
enarha, line 268:
```suggestion
3. The new pod starts, cache syncs, and takes ownership of assigned bucket
```

docs/horizontal-scaling.md
```diff
@@ -0,0 +1,291 @@
...
+4. Reconciliations continue with at-least-once delivery guarantees
+
+**Downtime:**
+
+Individual resources may experience reconciliation delays during rolling updates (up to the requeue interval), but no reconciliations are lost.
+
+### DNS Re-Resolution Latency
+
+gRPC re-resolves DNS every 30 minutes by default. When scaling the API Deployment up or down, existing watcher connections may take up to 30 minutes to discover new endpoints or remove old ones.
```
enarha: Can you please share the reference to that source code/doc? Checking vendor/google.golang.org/grpc/internal/resolver/dns/dns_resolver.go I see `MinResolutionInterval = 30 * time.Second` and `ResolvingTimeout = 30 * time.Second` but nothing about retry timeout or 30 minutes.

docs/horizontal-scaling.md
```diff
@@ -0,0 +1,291 @@
...
+
+1. Update `spec.replicas` in the watcher StatefulSet
+2. Update `buckets` in the config-leader-election ConfigMap to match
+3. Apply the updated configuration
+
+Knative's leader election framework redistributes buckets to the remaining replicas. Resources are rebalanced automatically without manual intervention.
+
+**Safe scale-down:**
+
+Kubernetes deletes StatefulSet pods in reverse ordinal order (2, 1, 0). Knative promotes remaining replicas to take ownership of orphaned buckets via informer resync. No reconciliations are lost during scale-down.
```
enarha: I believe this is wrong, see my comment below.

cmd/watcher/main.go
```diff
@@ -208,9 +216,7 @@ func connectToAPIServer(ctx context.Context, apiAddr string, authMode string) (*
 	}
 
 	log.Printf("dialing %s...\n", apiAddr)
-	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
-	defer cancel()
-	return grpc.DialContext(ctx, apiAddr, opts...)
+	return grpc.NewClient(apiAddr, opts...)
```
enarha, line 228: This is a significant change of behavior, because of the removal of `grpc.WithBlock()` which is not supported with `grpc.NewClient`. The client is no longer blocking, instead it returns immediately and relies on lazy connection establishment. I'm not against the change, but it should be very well documented both as part of the commit message (both the actual commit message and PR description) and the release notes (part of the PR description).

test/e2e/01-install-ha.sh
```diff
@@ -0,0 +1,85 @@
...
+
+export KO_DOCKER_REPO=${KO_DOCKER_REPO:-"kind.local"}
+export KIND_CLUSTER_NAME=${KIND_CLUSTER_NAME:-"tekton-results"}
+export SA_TOKEN_PATH=${SA_TOKEN_PATH:-"/tmp/tekton-results/tokens"}
+export SSL_CERT_PATH=${SSL_CERT_PATH:="/tmp/tekton-results/ssl"}
+
+ROOT="$(git rev-parse --show-toplevel)"
+
+echo "Installing Tekton Pipelines..."
+TEKTON_PIPELINE_CONFIG=${TEKTON_PIPELINE_CONFIG:-"https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml"}
```
enarha: `storage.googleapis.com` is the old storage and will result the installation of very old version of Tekton Pipeline. `test/e2e/01-install.sh` has the actual updated host.

test/e2e/01-install-ha.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
```
enarha: There is a lot of duplication (and some introduced errors) between this newly added script and the existing one test/e2e/01-install.sh. here is an Ai generated optimization idea. Please verify before using.

## Script duplication: `01-install.sh`, `01-install-ha.sh`, `generate-tls-cert.sh`

The three scripts share most of their logic. `01-install-ha.sh` is a near-complete copy of `01-install.sh` with only a few differences:

1. Extra SAN in the TLS cert (`api-service-headless`)
2. Different kustomize overlay (`kustomize-ha/` vs `kustomize/`)
3. Different pod wait strategy (label-based vs Deployment-based)

`generate-tls-cert.sh` duplicates the OpenSSL + LibreSSL cert generation a third time.

### Suggestion

Collapse into a single `01-install.sh` that accepts a mode parameter:

```bash
#!/bin/bash
set -e

MODE="${1:-standard}" # "standard" or "ha"

export KO_DOCKER_REPO=${KO_DOCKER_REPO:-"kind.local"}
export KIND_CLUSTER_NAME=${KIND_CLUSTER_NAME:-"tekton-results"}
export SA_TOKEN_PATH=${SA_TOKEN_PATH:-"/tmp/tekton-results/tokens"}
export SSL_CERT_PATH=${SSL_CERT_PATH:="/tmp/tekton-results/ssl"}

ROOT="$(git rev-parse --show-toplevel)"

# --- Install Tekton Pipelines (shared) ---
TEKTON_PIPELINE_CONFIG=${TEKTON_PIPELINE_CONFIG:-"https://infra.tekton.dev/tekton-releases/pipeline/latest/release.yaml"}
kubectl apply --filename "${TEKTON_PIPELINE_CONFIG}"
kubectl wait --for=condition=ready pod -l app=tekton-pipelines-controller -n tekton-pipelines --timeout=120s
kubectl wait --for=condition=ready pod -l app=tekton-pipelines-webhook -n tekton-pipelines --timeout=120s

# --- DB secret (shared) ---
kubectl create secret generic tekton-results-postgres \
    --namespace="tekton-pipelines" \
    --from-literal=POSTGRES_USER=postgres \
    --from-literal=POSTGRES_PASSWORD="$(openssl rand -base64 20)" || true

# --- TLS cert generation — delegate to the shared helper ---
export NAMESPACE="tekton-pipelines"
export OUTPUT_DIR="${SSL_CERT_PATH}"
export SECRET_NAME="tekton-results-tls"

if [ "$MODE" = "ha" ]; then
    # The generate-tls-cert.sh script already includes both the regular
    # and headless service names in the SAN.
    "${ROOT}/config/components/horizontal-scaling/generate-tls-cert.sh"
else
    # Standard mode: single SAN only.
    # Could extend generate-tls-cert.sh to accept EXTRA_SANS, or
    # keep the inline openssl call for the simple case.
    mkdir -p "${SSL_CERT_PATH}"
    altNames="DNS:tekton-results-api-service.tekton-pipelines.svc.cluster.local"
    if [ "${SSL_INCLUDE_LOCALHOST:-false}" = "true" ]; then
        altNames+=",DNS:localhost"
    fi
    openssl req -x509 \
        -newkey rsa:4096 \
        -keyout "${SSL_CERT_PATH}/key.pem" \
        -out "${SSL_CERT_PATH}/cert.pem" \
        -days 365 -nodes \
        -subj "/CN=tekton-results-api-service.tekton-pipelines.svc.cluster.local" \
        -addext "subjectAltName = ${altNames}"
    kubectl create secret tls -n tekton-pipelines tekton-results-tls \
        --cert="${SSL_CERT_PATH}/cert.pem" \
        --key="${SSL_CERT_PATH}/key.pem" || true
fi

# --- Deploy (mode-specific overlay) ---
if [ "$MODE" = "ha" ]; then
    KUSTOMIZE_DIR="${ROOT}/test/e2e/kustomize-ha"
else
    KUSTOMIZE_DIR="${ROOT}/test/e2e/kustomize"
fi
extra_ko_params="linux/$(go env GOARCH)"
kubectl kustomize "$KUSTOMIZE_DIR" | ko apply --platform="$extra_ko_params" -f -

# --- SA tokens (shared) ---
mkdir -p "${SA_TOKEN_PATH}"
for sa in all-namespaces-read-access single-namespace-read-access all-namespaces-admin-access all-namespaces-impersonate-access; do
    kubectl create token "$sa" > "${SA_TOKEN_PATH}/${sa}"
done

# --- Wait for pods (shared, label-based works for both modes) ---
kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=tekton-results-api -n tekton-pipelines --timeout=300s
kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=tekton-results-watcher -n tekton-pipelines --timeout=300s
kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=tekton-results-postgres -n tekton-pipelines --timeout=120s
```

Then update `e2e.sh` to call:
```bash
if [ "$1" == "--ha" ]; then
    ${REPO}/test/e2e/01-install.sh ha
else
    ${REPO}/test/e2e/01-install.sh
fi
```

To fully eliminate the third copy, `generate-tls-cert.sh` could be extended to support an `EXTRA_SANS` env var so it handles both the single-SAN and multi-SAN cases, making it usable as the sole cert generation entry point for all paths.

enarha (commented):
Thank you for the PR. Few comments, nothing major.
 
Please squash the commits. There is no really need to keep piling them for every change, unless you really want to split the PR into multiple commits encapsulating different functionality. Even when you squash, the diff is visible in GitHub if someone is interested in what changed between two pushes.

Update the actual commit message as well as the PR description. See my comment below, the PR description should also include release note as well as kind label (it should be `/kind feature` in this case).

Update the doc with information how the metrics should be scraped when using that ha mode. Something like `When using horizontal scaling, configure Prometheus to scrape individual pods (via PodMonitor or targeting the headless service) rather than the ClusterIP service. The ClusterIP only hits one random pod per scrape, giving an incomplete view. The headless services already expose the metrics port (9090), so they can be used as scrape targets directly.`.

docs/horizontal-scaling.md
olegbet: Seems like this statement is also not correct, I'll update the documentation.

test/e2e/01-install-ha.sh
```diff
@@ -0,0 +1,85 @@
...
+kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=tekton-results-api -n tekton-pipelines --timeout=300s
+kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=tekton-results-watcher -n tekton-pipelines --timeout=300s
+kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=tekton-results-postgres -n tekton-pipelines --timeout=120s
+
+echo ""
+echo "Installation complete!"
+echo ""
+kubectl get pods -n tekton-pipelines | grep tekton-results
+echo ""
+echo "Run E2E tests: go test -v --tags=e2e,e2e_ha -run TestHorizontalScaling ."
```
enarha: Run E2E tests: go test -v --tags=e2e,e2e_ha -run TestHorizontalScaling .
[ results] -> go test -v --tags=e2e,e2e_ha -run TestHorizontalScaling .
# .
no Go files in /tmp/results
FAIL    . [setup failed]
FAIL

that should be `"Run E2E tests: go test -v --tags=e2e,e2e_ha -run TestHorizontalScaling ./..."` or something stricter if you want to limit it to specific directory.

enarha:
/hold
something went wrong with that rebase

test/e2e/01-install-ha.sh
olegbet: Updated

test/e2e/01-install-ha.sh
olegbet: Refactored the scripts

cmd/watcher/main.go
enarha: I tested this manually and it created a bigger issue. The logs functionality is completely skipped because grpc.NewClient returns immediately and the DNS resolution is not complete yet. The message in the log `Unable to inject logs client, logs will not be stored: rpc error: code = Unavailable desc = no children to pick from` indicates that. I tested a fix which seems to be working:
```
diff --git a/cmd/watcher/main.go b/cmd/watcher/main.go
index 6b99887..624086e 100644
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -191,6 +191,7 @@ func connectToAPIServer(apiAddr string, authMode string) (*grpc.ClientConn, erro
        opts := []grpc.DialOption{
                grpc.WithDefaultServiceConfig(defaultServiceConfig),
                grpc.WithConnectParams(connectParams),
+               grpc.WithDefaultCallOptions(grpc.WaitForReady(true)),
        }
        // Add in additional credentials to requests if desired.
        switch authMode {
```
this is also what https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md recommends.

config/components/horizontal-scaling/watcher-statefulset.yaml
```diff
@@ -0,0 +1,107 @@
...
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
+              labelSelector:
+                matchLabels:
+                  app.kubernetes.io/name: tekton-results-watcher
+              topologyKey: kubernetes.io/hostname
+            weight: 100
+      serviceAccountName: watcher
+      containers:
```
adityavshinde, line 53: Missing liveness/readiness probes on the watcher StatefulSet:
We should define a livenessProbe or readinessProbe. Knative's 'shareadmin' already serves /heath and /readiness on port 8080 inside the watcher, but the StatefulSet spec never tells Kubernetes to check them. In the HA setup with bucket sharding, this means if one watcher replica gets stuck, K8s will keep reporting it as "Ready 1/1". That replica's bucket of TR/PR will silently stop being reconciled ( no alert or detection ).
IMO we can fix this by adding livenessProbe or readinessProbe to the container spec ( port 8080 also needs to be added )

enarha: While I agree with that, it's a pre-existing state and it applies to the Deployment as well (when we deploy as single replica). So IMO it makes sense to fix that in a separate PR which fixes the issue for both Deployment and StatefulSet. We can open a separate GitHub issue for it, so it's not forgotten.

config/components/horizontal-scaling/watcher-statefulset.yaml
```diff
@@ -0,0 +1,107 @@
...
+            - name: CONFIG_LOGGING_NAME
+              value: tekton-results-config-logging
+            - name: CONFIG_LEADERELECTION_NAME
+              value: tekton-results-config-leader-election
+            - name: CONFIG_OBSERVABILITY_NAME
+              value: tekton-results-config-observability
+            - name: METRICS_DOMAIN
+              value: tekton.dev/results
+            - name: TEKTON_RESULTS_API_SERVICE
+              value: dns:///tekton-results-api-service-headless.tekton-pipelines.svc.cluster.local:8080
```
adityavshinde, line 75: The TEKTON_RESULTS_API_SERVICE env var hardcodes tekton-pipelines in the DNS name. The base Deployment has the same hardcoding, but it gets patched by the operator at deploy time. Since this StatefulSet is a new resource, the operator would also need to be updated to patch it for non-default namespaces.

config/components/horizontal-scaling/watcher-statefulset.yaml
adityavshinde: Yes, that makes sense to have a different PR for this.

adityavshinde:
PodDisruptionBudget (PDB) for HA components can be added:
This PR has 3 replica HA for both the watcher (StatefulSet) and API server (Deployment), but there is no PodDisruptionBudget to protect availability during voluntary disruptions like node drains or cluster upgrades.
Without a PDB, kubectl drain can evict all 3 watcher pods or all 3 API pods simultaneously if they happen to be co-located. With StatefulSet ordinals, evicted pods' buckets have no failover, so a drain that hits 2 pods means 2/3 of reconciliation stops until the pods are rescheduled.
This is not a blocker, but this can be a better fit for HA component.
We can add a policy under config/components/horizontal-scaling/

olegbet:
> PodDisruptionBudget (PDB) for HA components can be added:
> This PR has 3 replica HA for both the watcher (StatefulSet) and API server (Deployment), but there is no PodDisruptionBudget to protect availability during voluntary disruptions like node drains or cluster upgrades.
> Without a PDB, kubectl drain can evict all 3 watcher pods or all 3 API pods simultaneously if they happen to be co-located. With StatefulSet ordinals, evicted pods' buckets have no failover, so a drain that hits 2 pods means 2/3 of reconciliation stops until the pods are rescheduled.
> This is not a blocker, but this can a better fit for HA component.
> We can add a policy under config/components/horizontal-scaling/

Added PodDisruptionBudget to the watcher StatefulSet.

config/components/horizontal-scaling/watcher-statefulset.yaml
enarha: Good call. I'll add to that, that the operator should not just patch, but ultimately manage that new Service object and the Route associated with it. Here we do not have that capability to change dynamically the type of deployment (single replicas vs scaleable deployments). The work on the operator side can be managed as a separate issue on the operator repo. We need that PR merged first and the next version of Results including that change to be referenced by the operator because of the change in the Watcher main file.

config/components/horizontal-scaling/watcher-statefulset.yaml
adityavshinde: I have created an issue for the above: https://github.com/tektoncd/results/issues/1466

test/e2e/e2e.sh
```diff
@@ -37,25 +37,32 @@ main() {
...
+        kubectl delete pod $(kubectl get pod -o=name -n tekton-pipelines | grep tekton-results-api | sed "s/^.\{4\}//") -n tekton-pipelines
+        kubectl wait deployment "tekton-results-api" --namespace="tekton-pipelines" --for="condition=available" --timeout="120s"
+        kubectl delete pod $(kubectl get pod -o=name -n tekton-pipelines | grep tekton-results-watcher | sed "s/^.\{4\}//") -n tekton-pipelines
+        kubectl wait deployment "tekton-results-watcher" --namespace="tekton-pipelines" --for="condition=available" --timeout="120s"
+        go test -v -count=1 --tags=e2e,gcs $(go list --tags=e2e ${REPO}/test/e2e/... | grep -v /client) -run TestGCSLog
+    fi
+
 }
 
 main
```
adityavshinde: `main` on line 68 is called without `"$@"`, so the function never receives the "ha" argument from the command line. `$#` inside `main()` is always 0 and the HA branch at line 42 is dead code. This should be main `"$@"`

test/e2e/e2e_ha_test.go
```diff
@@ -0,0 +1,462 @@
...
+				stream.Close()
+				if err != nil {
+					t.Errorf("Error reading logs for watcher pod %s: %v", pod.Name, err)
+					continue
+				}
+
+				logs := string(logsBytes)
+
+				// Check which TaskRuns appear in this pod's logs
+				for _, trName := range taskRunNames {
```
adityavshinde: Even if the `main "$@"` fix is applied to make the HA branch reachable in e2e.sh, the HA path (lines 42-44) doesn't patch `tekton-results-config-logging` to set l`oglevel.watcher: debug`, unlike the standard path (lines 49-52). The watcher-side check sub test in e2e_ha_test.go searches pod logs for "knative.dev/key" entries to verify bucket sharding, but these entries are only emitted at debug level. At the default info level, the watcher processes TaskRuns correctly but never logs the key entries, so the test finds zero matches and reports a false failure.

test/e2e/e2e.sh
olegbet: Adressed

test/e2e/e2e_ha_test.go
olegbet: Addressed

adityavshinde:
I have reviewed this PR and the changes looks good to me. /lgtm

docs/horizontal-scaling.md
enarha: this is still not updated

docs/horizontal-scaling.md
```diff
@@ -0,0 +1,311 @@
...
+The watcher uses `grpc.NewClient` (introduced in gRPC-Go 1.58) instead of the deprecated `grpc.DialContext`.
+
+**Key differences:**
+
+| `grpc.DialContext` (deprecated) | `grpc.NewClient` (current) |
+|--------------------------------|---------------------------|
+| Blocking by default | Non-blocking - returns immediately |
+| Accepts `context.Context` | No context parameter |
+| Supports `grpc.WithBlock()` | Does not support `grpc.WithBlock()` |
+| Connection established during dial | Connections established lazily on first RPC |
```
enarha: This is more of a implementation detail, while I see this doc as a customer focused documentation. IMO we can just drop that.
```suggestion
```

test/e2e/e2e.sh
```diff
@@ -37,25 +37,36 @@ main() {
...
+        kubectl wait deployment "tekton-results-api" --namespace="tekton-pipelines" --for="condition=available" --timeout="120s"
+        kubectl delete pod $(kubectl get pod -o=name -n tekton-pipelines | grep tekton-results-watcher | sed "s/^.\{4\}//") -n tekton-pipelines
+        kubectl wait deployment "tekton-results-watcher" --namespace="tekton-pipelines" --for="condition=available" --timeout="120s"
+        go test -v -count=1 --tags=e2e,gcs $(go list --tags=e2e ${REPO}/test/e2e/... | grep -v /client) -run TestGCSLog
+    fi
+
 }
 
-main
+main $@
```
enarha:
```suggestion
main "$@"
```

docs/horizontal-scaling.md
```diff
@@ -0,0 +1,311 @@
...
+
+**How it works:**
+
+1. Knative's controller framework partitions the reconciliation key space into buckets using consistent hashing
+2. Each StatefulSet replica is assigned one or more buckets based on its ordinal (derived from pod name)
+3. When a watcher observes a TaskRun, PipelineRun, or CustomRun, it computes the bucket hash from the resource's namespace/name
+4. The watcher only reconciles the resource if the computed bucket matches one of its assigned buckets
+5. Resources are deterministically assigned to exactly one watcher, eliminating duplicate reconciliation
+
+Note: Deprecated blocking grpc.DialContext client connection function has been replaced with newer grpc.NewClient function. grpc.WithBlock is uncompatible with the new scaling functionality and has been removed and replaced with grpc.WithConnectParams(connectParams) to use new scaling parameters;
```
enarha: typo
```suggestion
Note: Deprecated blocking grpc.DialContext client connection function has been replaced with newer grpc.NewClient function. grpc.WithBlock is incompatible with the new scaling functionality and has been removed and replaced with grpc.WithConnectParams(connectParams) to use new scaling parameters;
```

test/e2e/e2e.sh
adityavshinde: @enarha Thanks for pointing this out, I missed it. Yes it should be `main "$@"`

olegbet: I tried this locally and this works both ways but thanks anyways, I'll update it.

olegbet: fixed

docs/horizontal-scaling.md
olegbet: fixed

docs/horizontal-scaling.md
olegbet: dropped

docs/horizontal-scaling.md
olegbet: updated this part

docs/horizontal-scaling.md
olegbet: updated in the last commit

test/e2e/e2e.sh
enarha: > I tried this locally and this works both ways but thanks anyways, I'll update it.

It works with the existing `ha` parameter, but that could break when the shell performs word splitting and glob expansion with other parameters we add to it. This is a shell correctness issue.

enarha:
/unhold
/approve
lgtm on behalf of @adityavshinde 
/lgtm

PR #1442, opened 2026-08-17
feat(watcher): filter TaskRun/PipelineRun by spec.managedBy

khrm:
Add filtering so the Results watcher only processes TaskRuns and
  PipelineRuns whose `spec.managedBy` value is allowed. By default,
  `tekton.dev/pipeline` is always accepted, as are runs with unset,
  empty, or whitespace-only `managedBy`. Users can configure additional
  allowed values via the `--managed_by_values` flag.

  The filter is applied in two places:
  - **Informer level**: `cache.FilteringResourceEventHandler` prevents enqueuing runs with disallowed `managedBy` on normal watch events.
  - **Leader promotion**: `PromoteFilterFunc` in `controller.Options` prevents re-enqueuing those same runs when a replica becomes leader.

  Tombstone (`DeletedFinalStateUnknown`) objects are unwrapped before the type assertion so that delete events are never silently dropped.

  CustomRuns are not filtered because the `CustomRun` type does not have a `spec.managedBy` field.

  Similar to [tektoncd/chains#1882](https://github.com/tektoncd/chains/pull/1882).

# Changes

# Submitter Checklist

# Release Notes

```release-note
The Results Watcher now supports filtering TaskRuns and PipelineRuns by `spec.managedBy`.
  Use the new `--managed_by_values` flag to specify which controllers' runs the Watcher should
  process. `tekton.dev/pipeline` and unset values are always accepted. Runs with disallowed
  `managedBy` values are ignored at the informer, reconciler, and leader election layers.
  If the allowlist is later shrunk, existing finalizers are cleaned up on deletion, including
  legacy merge-patch finalizers from pre-SSA versions.
```

### DIFF
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -41,6 +41,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/credentials/oauth"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/transport"
@@ -83,6 +84,7 @@ var (
 	disableStoringIncompleteRuns = flag.Bool("disable_storing_incomplete_runs", false, "If enabled, Runs will not be stored until they are complete. If disabled, Runs will be initially stored upon creation and continuously upserted until after they're deleted")
 	storeDeadline                = flag.Duration("store_deadline", 10*time.Minute, "How long to wait for storing the PipelineRun and TaskRun resources before aborting and clearing the finalizer in case of delete event")
 	forwardBuffer                = flag.Duration("forward_buffer", 150*time.Second, "This determines duration since completion time of TaskRun to wait for forwarder to finish")
+	managedByValues              = flag.String("managed_by_values", "", "Comma-separated list of additional spec.managedBy values the watcher will process. Runs with unset, empty, whitespace-only or \"tekton.dev/pipeline\" managedBy values are always accepted.")
 )
 
 func main() {
@@ -136,9 +138,11 @@ func main() {
 		SummaryLabels:                *summaryLabels,
 		SummaryAnnotations:           *summaryAnnotations,
 		DisableStoringIncompleteRuns: *disableStoringIncompleteRuns,
+		AllowedManagedByValues:       reconciler.ParseManagedByValues(*managedByValues),
 	}
 
 	log.Printf("dynamic reconcile timeout %s and update log timeout is %s", cfg.DynamicReconcileTimeout.String(), cfg.UpdateLogTimeout.String())
+	log.Printf("managed_by_values: %v", sets.List(cfg.AllowedManagedByValues))
 
 	if selector := *labelSelector; selector != "" {
 		if err := cfg.SetLabelSelector(selector); err != nil {
@@ -169,7 +173,8 @@ func main() {
 		k8scfg.Burst = *burst * len(ctors)
 	}
 
-	sharedmain.MainWithConfig(injection.WithNamespaceScope(ctx, *namespace), "watcher", k8scfg, ctors...,
+	sharedmain.MainWithConfig(
+		injection.WithNamespaceScope(ctx, *namespace), "watcher", k8scfg, ctors...,
 	)
 }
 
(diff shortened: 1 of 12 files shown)


khrm (commented):
/kind feature

cmd/watcher/main.go
```diff
@@ -83,6 +84,7 @@ var (
 	disableStoringIncompleteRuns = flag.Bool("disable_storing_incomplete_runs", false, "If enabled, Runs will not be stored until they are complete. If disabled, Runs will be initially stored upon creation and continuously upserted until after they're deleted")
 	storeDeadline                = flag.Duration("store_deadline", 10*time.Minute, "How long to wait for storing the PipelineRun and TaskRun resources before aborting and clearing the finalizer in case of delete event")
 	forwardBuffer                = flag.Duration("forward_buffer", 150*time.Second, "This determines duration since completion time of TaskRun to wait for forwarder to finish")
+	managedByValues              = flag.String("managed_by_values", "", "Comma-separated list of additional spec.managedBy values the watcher will process. \"tekton.dev/pipeline\" is always accepted and cannot be removed. Runs with unset, empty, or whitespace-only managedBy are also always accepted.")
```
enarha:
```suggestion
	managedByValues              = flag.String("managed_by_values", "", "Comma-separated list of additional spec.managedBy values the watcher will process. Runs with unset, empty, whitespace-only or \"tekton.dev/pipeline\" managedBy values are always accepted.")
```

pkg/watcher/reconciler/pipelinerun/reconciler.go
```diff
@@ -66,15 +66,22 @@ type Reconciler struct {
...
+	_ pipelinerunreconciler.Finalizer = (*Reconciler)(nil)
+)
 
 // ReconcileKind makes new watcher reconcile cycle to handle PipelineRun.
 func (r *Reconciler) ReconcileKind(ctx context.Context, pr *pipelinev1.PipelineRun) knativereconciler.Event {
 	logger := logging.FromContext(ctx).With(zap.String("results.tekton.dev/kind", "PipelineRun"))
 
 	logger.Infof("Initiating reconciliation for PipelineRun '%s/%s'", pr.Namespace, pr.Name)
 
+	if !reconciler.IsManagedByAllowed(pr.Spec.ManagedBy, r.cfg.AllowedManagedByValues) {
```
enarha, line 80: Please add a unit test for this case and same for taskrun. It should go into TestReconcile(), test case `disallowed managedBy - skip`.

pkg/watcher/reconciler/pipelinerun/reconciler.go
```diff
@@ -200,6 +207,20 @@ func isCustomRunMarkedAsReadyForDeletion(customRun *pipelinev1beta1.CustomRun) b
 // that we see flowing through the system.  If we don't add a finalizer, it could
 // get cleaned up before we see the final state and store it.
 func (r *Reconciler) FinalizeKind(ctx context.Context, pr *pipelinev1.PipelineRun) knativereconciler.Event {
+	if !reconciler.IsManagedByAllowed(pr.Spec.ManagedBy, r.cfg.AllowedManagedByValues) {
```
enarha, line 210: It would be nice to add unit test for that. The existing TestFinalize() test does not call FinalizeKind, but finalize(), so it'll skip that check. Instead we'll need TestFinalizeKindDisallowedManagedBy specific test which only checks that functionality.

pkg/watcher/reconciler/pipelinerun/reconciler.go
khrm: Addressed this.

pkg/watcher/reconciler/pipelinerun/reconciler.go
khrm: Addressed this also.

enarha:
Please add a release note for this feature

/approve

pkg/watcher/reconciler/filter.go
```diff
@@ -0,0 +1,111 @@
...
+	return obj
+}
+
+// PipelineRunFilterFunc returns a filter function for PipelineRuns
+// that checks whether the run's spec.managedBy value is allowed.
+// Tombstone (DeletedFinalStateUnknown) objects are unwrapped before
+// the type assertion so that delete events are never silently dropped.
+// Objects that already carry a Results finalizer are always passed
+// through so the finalizer can be removed even if the allowlist
+// has been shrunk since the finalizer was added.
```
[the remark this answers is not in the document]
khrm: Fixed the comment.

divyansh42:
@khrm could you please add release notes?

divyansh42 (commented):
/lgtm

PR #1453, opened 2026-08-19
Make CI workflows reusable via workflow_call for cross-branch reuse

divyansh42:
## Summary

Adds an `on: workflow_call:` trigger to `codeql.yml`, `zizmor.yaml`, `golangci-lint.yaml`, `presubmit-ci.yaml`, and `go-coverage.yml` so that active release branches can call these canonical workflow definitions directly instead of maintaining full duplicated copies.

This is the first of a set of companion PRs against each active release branch (`release-v0.17.x`, `release-v0.18.x`, `release-v0.19.x`, `release-v0.20.x`) that replace their duplicated workflow files with thin wrappers using:

```yaml
jobs:
  some-job:
    uses: tektoncd/results/.github/workflows/<file>@main
```

Editing workflow logic on `main` then automatically applies everywhere, instead of needing to hand-edit each release branch.

Existing `push`/`pull_request`/`schedule` triggers on `main` are unchanged, so `main`'s own CI behavior is unaffected by this change — it purely adds the ability for other workflows/branches to call these as reusable workflows.

## Test plan

- [ ] Confirm CI still runs as before on this PR (push/pull_request triggers unchanged)
- [ ] Merge this PR before merging the companion release-branch PRs, since they reference `uses: tektoncd/results/.github/workflows/<file>@main`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -19,6 +19,7 @@ on:
     branches: ["main"]
   schedule:
     - cron: '38 3 * * 3'
+  workflow_call: {}
 
 jobs:
   analyze:
--- a/.github/workflows/go-coverage.yml
+++ b/.github/workflows/go-coverage.yml
@@ -6,7 +6,8 @@ permissions:
 on:
   push:
     branches: ["main"]
-  workflow_dispatch:
+  workflow_dispatch: {}
+  workflow_call: {}
 
 concurrency:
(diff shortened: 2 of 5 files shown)


khrm (commented):
/approve
/lgtm

enarha (commented):
/lgtm

PR #1477, opened 2026-09-02
feat(e2e): add Postgres DB layer e2e test suite

adityavshinde:
# Changes

Add Postgres-specific e2e test suite (`test/e2e/db/`) that exercises database
behavior the SQLite-backed unit tests cannot reach: `pgconn.PgError` SQLSTATE
error mapping, `jsonb` column types and operators, CEL -> SQL filter/sort/pagination
on real Postgres, FK cascade, and schema correctness.

Tests run against the deployed API server in the kind cluster and its live
Postgres instance via gRPC - not an in-process server or throwaway database.

**What's included:**
- Error mapping tests (unique violation, FK violation, stale etag, NotFound, cascade delete)
- Schema validation (column types, PKs, unique indexes, FKs, `jsonb @>`)
- Lister contract tests (filter, order, pagination with `pageSize≥5`, ListRecords)
- Placeholder stubs with `t.Skip` for future stories (labels, migrations, relationships, retention)
- Shared client config extracted to `test/e2e/client/config.go` (eliminates duplication between `e2e_test.go` and `db_test.go`)
- CI wiring in `e2e.sh`: port-forward on `15432`, `DB_URL` env var, password suppressed from traces, cleanup on failure
- `test/e2e/README.md` and `test/e2e/db/README.md` updated/created

/kind feature

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/client/config.go
+++ b/test/e2e/client/config.go
@@ -1,0 +1,108 @@
+package client
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"k8s.io/client-go/transport"
+)
+
+// Default e2e environment values. These match the paths used by
+// test/e2e/01-install.sh and the kind cluster setup scripts.
+const (
+	DefaultServerName    = "tekton-results-api-service.tekton-pipelines.svc.cluster.local"
+	DefaultServerAddress = "https://localhost:8080"
+	DefaultCertFileName  = "tekton-results-cert.pem"
+	DefaultCertPath      = "/tmp/tekton-results/ssl"
+	DefaultTokenPath     = "/tmp/tekton-results/tokens" //nolint:gosec // Not a credential; directory path for SA token files.
+
+	AdminTokenFile = "all-namespaces-admin-access"
+	ReadTokenFile  = "all-namespaces-read-access"
+)
+
+// EnvConfig holds the resolved paths and addresses for the e2e test environment.
+type EnvConfig struct {
+	CertFile      string
+	TokenPath     string
+	ServerName    string
+	ServerAddress string
+}
+
+// NewEnvConfig reads the standard e2e environment variables and falls back to
+// defaults. Both the main e2e suite and the db sub-suite share this to avoid
+// duplicating the env-var-to-default resolution logic.
+func NewEnvConfig() EnvConfig {
+	certPath := EnvOrDefault("SSL_CERT_PATH", DefaultCertPath)
+	certFileName := EnvOrDefault("CERT_FILE_NAME", DefaultCertFileName)
+	return EnvConfig{
+		CertFile:      path.Join(certPath, certFileName),
+		TokenPath:     EnvOrDefault("SA_TOKEN_PATH", DefaultTokenPath),
+		ServerName:    EnvOrDefault("API_SERVER_NAME", DefaultServerName),
+		ServerAddress: EnvOrDefault("API_SERVER_ADDR", DefaultServerAddress),
+	}
+}
+
+// TokenFile returns the full path for a given token file name under the
+// resolved token directory.
+func (c EnvConfig) TokenFile(name string) string {
+	return path.Join(c.TokenPath, name)
+}
+
+// NewGRPCClientFromConfig creates a GRPCClient using the shared EnvConfig and
+// the specified token file name.
+func NewGRPCClientFromConfig(cfg EnvConfig, tokenFileName string, impersonationConfig *transport.ImpersonationConfig) (GRPCClient, error) {
+	if impersonationConfig == nil {
+		impersonationConfig = &transport.ImpersonationConfig{}
+	}
+
+	transportCreds, err := credentials.NewClientTLSFromFile(cfg.CertFile, cfg.ServerName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load TLS credentials from %s: %w", cfg.CertFile, err)
+	}
+
+	opts := []grpc.DialOption{
+		grpc.WithBlock(), //nolint:staticcheck
+		grpc.WithTransportCredentials(transportCreds),
+		grpc.WithDefaultCallOptions(grpc.PerRPCCredentials(&CustomCredentials{
+			TokenSource:         transport.NewCachedFileTokenSource(cfg.TokenFile(tokenFileName)),
+			ImpersonationConfig: impersonationConfig,
+		})),
+	}
+
+	return NewGRPCClient(cfg.ServerAddress, opts...)
+}
+
+// NewRESTClientFromConfig creates a RESTClient using the shared EnvConfig and
+// the specified token file name.
+func NewRESTClientFromConfig(cfg EnvConfig, tokenFileName string, impersonationConfig *transport.ImpersonationConfig) (RESTClient, error) {
+	if impersonationConfig == nil {
+		impersonationConfig = &transport.ImpersonationConfig{}
+	}
+
+	if _, err := credentials.NewClientTLSFromFile(cfg.CertFile, cfg.ServerName); err != nil {
+		return nil, fmt.Errorf("failed to verify TLS cert from %s: %w", cfg.CertFile, err)
+	}
+
+	restConfig := &transport.Config{
+		TLS: transport.TLSConfig{
+			CAFile:     cfg.CertFile,
+			ServerName: cfg.ServerName,
+		},
+		BearerTokenFile: cfg.TokenFile(tokenFileName),
+		Impersonate:     *impersonationConfig,
+	}
+
+	return NewRESTClient(cfg.ServerAddress, WithConfig(restConfig))
+}
+
+// EnvOrDefault returns the value of the environment variable named by key,
+// or fallback if the variable is empty or unset.
+func EnvOrDefault(key, fallback string) string {
+	if v := os.Getenv(key); v != "" {
+		return v
+	}
+	return fallback
+}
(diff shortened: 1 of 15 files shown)


test/e2e/db/lister_test.go
```diff
@@ -0,0 +1,444 @@
...
+		})
+		if err != nil {
+			t.Fatalf("page %d: ListResults failed: %v", page, err)
+		}
+
+		for _, r := range resp.Results {
+			if r.Summary.Status != pb.RecordSummary_SUCCESS {
+				t.Errorf("filter leaked non-SUCCESS result: %s (status=%v)", r.Name, r.Summary.Status)
+			}
+			successNames = append(successNames, r.Name)
```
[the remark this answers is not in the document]
adityavshinde: Resolved

test/e2e/db/schema_test.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		t.Fatalf("CreateResult failed: %v", err)
+	}
+	t.Cleanup(func() { deleteResult(t, res.GetName()) })
+
+	var count int64
+	err = rawDB.Raw(
+		`SELECT count(*) FROM results WHERE annotations @> ?`,
+		`{"env":"production"}`,
+	).Scan(&count).Error
```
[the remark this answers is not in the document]
adityavshinde: Resolved

test/e2e/client/config.go
```diff
@@ -0,0 +1,108 @@
+package client
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"google.golang.org/grpc"
```
khrm, line 8: Did you run gofumt here? Is it github or somehow this is rendering weirdly for me.

khrm: It was github issue.

test/e2e/db/lister_test.go
```diff
@@ -0,0 +1,461 @@
...
+
+	resp, err := readClient.ListResults(defaultCtx(), &pb.ListResultsRequest{
+		Parent:  parent,
+		OrderBy: "create_time asc",
+	})
+	if err != nil {
+		t.Fatalf("ListResults with ORDER BY ASC failed: %v", err)
+	}
+
+	for i := 1; i < len(resp.Results); i++ {
```
khrm, line 244: We should also assert the length of the results returned.

test/e2e/e2e.sh
```diff
@@ -28,6 +28,47 @@ cleanup() {
 
 trap cleanup EXIT
 
+run_db_tests() {
+    REPO="$1"
+    local LOCAL_PG_PORT=15432
```
khrm: Instead of hardcoding this, we should have something like "${PG_E2E_PORT:-15432}"  Otherwise, only one test can be run, not parallel run with multiple PG instances..

test/e2e/db/error_mapping_test.go
```diff
@@ -0,0 +1,247 @@
...
+	ctx := context.Background()
+	const parent = "db-err-cascade"
+
+	res, err := adminClient.CreateResult(ctx, &pb.CreateResultRequest{
+		Parent: parent,
+		Result: &pb.Result{Name: parent + "/results/cascade-parent"},
+	})
+	if err != nil {
+		t.Fatalf("CreateResult failed: %v", err)
+	}
```
khrm, line 202: Why no cleanup here? If there's a failure, then there's leak. In case of pass, that cleanup can be a no ops.

test/e2e/db/README.md
```diff
@@ -0,0 +1,81 @@
...
+go test -v -count=1 -tags=e2e ./test/e2e/db/...
+```
+
+## Environment variables
+
+| Variable | Default | Description |
+|---|---|---|
+| `DB_URL` | (none) | Full Postgres DSN. When set, takes precedence over individual `POSTGRES_*` vars. The `e2e.sh` script constructs this automatically. |
+| `POSTGRES_HOST` | `localhost` | Fallback: Postgres host (via port-forward) |
+| `POSTGRES_PORT` | `15432` | Fallback: local forwarded port |
```
khrm: I see that you multiple vars here. Can you check whether you are using these var with default value in the script?

adityavshinde: The `POSTGRES_*` variables are used as fallbacks in `db_test.go`'s `buildDSN()` for manual runs when `DB_URL` is not set, the `e2e.sh` CI script always constructs `DB_URL` directly, so these individual vars are never used in the CI path.  I am mentioning this in  `README.md` file only.

test/e2e/e2e_test.go
```diff
@@ -19,102 +19,45 @@ package e2e
...
-	defaultServerAddress                        = "https://localhost:8080"
-	defaultCertFileName                         = "tekton-results-cert.pem"
-	allNamespacesReadAccessTokenFileName        = "all-namespaces-read-access"
-	singleNamespaceReadAccessTokenFileName      = "single-namespace-read-access"
-	allNamespacesAdminAccessTokenFileName       = "all-namespaces-admin-access"
-	allNamespacesImpersonateAccessTokenFileName = "all-namespaces-impersonate-access"
-	defaultCertPath                             = "/tmp/tekton-results/ssl"
-	defaultTokenPath                            = "/tmp/tekton-results/tokens"
-	defaultNamespace                            = "default"
+	allNamespacesReadAccessToken        = "all-namespaces-read-access"
```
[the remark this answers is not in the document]
adityavshinde: Resolved

test/e2e/db/error_mapping_test.go
adityavshinde: Resolved

test/e2e/e2e.sh
adityavshinde: Resolved

test/e2e/db/lister_test.go
adityavshinde: Resolved

test/e2e/client/config.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	restConfig := &transport.Config{
+		TLS: transport.TLSConfig{
+			CAFile:     cfg.CertFile,
+			ServerName: cfg.ServerName,
+		},
+		BearerTokenFile: cfg.TokenFile(tokenFileName),
+		Impersonate:     *impersonationConfig,
+	}
```
olegbet: The creds are created just to check that the certificate exists and get discarded later. Probably it can be done to be reused, something like:


```suggestion
caData, err := os.ReadFile(cfg.CertFile)
if err != nil {
      return nil, fmt.Errorf("failed to read TLS cert from %s: %w", cfg.CertFile, err)
}
if pool := x509.NewCertPool(); !pool.AppendCertsFromPEM(caData) {
      return nil, fmt.Errorf("no valid certificate found in %s", cfg.CertFile)
}

restConfig := &transport.Config{
      TLS: transport.TLSConfig{
              CAData:     caData,
              ServerName: cfg.ServerName,
      },
      BearerTokenFile: cfg.TokenFile(tokenFileName),
      Impersonate:     *impersonationConfig,
}
```

test/e2e/db/lister_test.go
```diff
@@ -0,0 +1,469 @@
...
+	const parentB = "db-list-wc-b"
+	t.Cleanup(func() {
+		cleanupParent(t, parentA)
+		cleanupParent(t, parentB)
+	})
+
+	seededA := seedResults(t, parentA, 3, pb.RecordSummary_SUCCESS)
+	seededB := seedResults(t, parentB, 3, pb.RecordSummary_FAILURE)
+
+	resp, err := readClient.ListResults(defaultCtx(), &pb.ListResultsRequest{
```
khrm: I think we need orderBy also. Otherwise, if we have more than 50 records due to earlier tests, this will fail.

khrm: Or we can use next page token but that's more complicated.

test/e2e/db/schema_test.go
```diff
@@ -0,0 +1,244 @@
...
+	var deleteRule string
+	err := rawDB.Raw(`
+		SELECT rc.delete_rule
+		FROM   information_schema.referential_constraints rc
+		JOIN   information_schema.table_constraints tc
+		  ON   rc.constraint_name = tc.constraint_name
+		  AND  rc.constraint_schema = tc.constraint_schema
+		JOIN   information_schema.constraint_table_usage ctu
+		  ON   rc.unique_constraint_name = ctu.constraint_name
+		  AND  rc.constraint_schema = ctu.constraint_schema
```
khrm: Is this line correct? ATM, this is working because both results and records schema live in public. Instead, we should have prolly: ```rc.unique_constraint_name = ctu.constraint_name```

test/e2e/db/db_test.go
```diff
@@ -0,0 +1,129 @@
+//go:build e2e
+// +build e2e
```
enarha: We do not need that, this is the older format of the tags used in Golang prior to 1.17. The same goes for all files in that package. `//go:build e2e` does the same thing.

test/e2e/db/db_test.go
```diff
@@ -0,0 +1,129 @@
...
+// A raw database connection is used only for schema introspection and to
+// trigger FK violations that the server's pre-checks would otherwise mask.
+//
+// Future stories (labels, metadata columns, golang-migrate migrations) add
+// their Postgres-specific tests to this package. See the stub files for
+// placeholders.
+//
+// Run:
+//
+//	go test -v -count=1 -tags=e2e ./test/e2e/db/...
```
enarha: this by itself fails without the existing kind environment. I'll suggest to remove it, instead such information should live in the README which complete steps of how to run all tests or a subset of them.

